PR #50, opened 2021-11-17
refactor and /image fix

JohnnyJumper:
I have moved part of the metadata endpoint logic outside to reuse in /image and /image-src

/image-src is a little weird as it was expecting the .svg type of image from the unstoppable cdn. I would assume that if we return a base64 when an nft image is attached is going to be the same behavior as returning a "path" from svg ?


JohnnyJumper:
Not ready. I found out that image-src is not displaying the image correctly. Working on a fix

JohnnyJumper:
Apparently, if the endpoint returns base64 string it won't automatically shows in the browser, when users place it inside the  tag. 
If it is a base64 string they will need to make a separate network call to get the string and then place the string inside the src attribute.
 If they will just use API link inside the src when base64 is returned, the  is not working properly. It is getting confused as with the HTTP protocol it expects the image bytes and not the base64 string encoded version. 
 
I updated the code a little to make sure we can control whether we want to return the base64 from nft or not. and not returning base64 inside the /image-src anymore.

DeRain:
@JohnnyJumper This is possible for all types of images supported by browsers. Check this guide and update the code accordingly. 

https://www.w3docs.com/snippets/html/how-to-display-base64-images-in-html.html

src/controllers/MetaDataController.ts
```diff
@@ -221,14 +206,27 @@ export class MetaDataController {
...
       return '';
     }
 
-    const imageData = await this.generateImageData(name, resolution);
-    return await pathThatSvg(imageData);
+    let socialPicture = await getSocialPicture({
+      domainName: name,
+      avatarRecord: records['social.picture.value'] || '',
+      ownerAddress: resolution.ownerAddress || '',
+      toBase64: false,
```
DeRain, line 222: All images could be converted to base64 and could be understood by browsers https://www.w3docs.com/snippets/html/how-to-display-base64-images-in-html.html

DeRain (changes requested):

src/controllers/MetaDataController.ts
DeRain: @JohnnyJumper there are a bunch of packages to solve this task. For example:
- https://www.npmjs.com/package/image-to-base64
- https://www.npmjs.com/package/node-base64-image

DeRain (changes requested):

JohnnyJumper:
@DeRain  Yes it is possibile for a browser to show the base64 encoded image, if and only if user will put the the base64 string inside the src attribute. However, image-src endpoint was designed to be placed inside the src attribute like so:
< img src = "https://metadata.unstoppabledomains.com/image-src/brad.crypto" /> In this case if the endpoint returns base64 string the browser will not show anything.
https://stackoverflow.com/questions/35828806/use-base64-string-from-url-in-src-tag-of-image/35829286

PR #54, opened 2021-11-19
Change metadata to return image URLs instead of base64

DancingAxolotl:
## Related story
https://www.pivotaltracker.com/story/show/180168262

## Changes
 - Changed the metadata endpoint to return image urls. Instead of base64 image data it returns a link to metadata `/image-src` endpoint which renders the image
 - Metadata endpoint doesn't pull image data, just checks if a custom image exists

## Notes
Should be merged after #50


DeRain (approved):
Great job!

DeRain (approved):

PR #59, opened 2021-12-21
Add parallel tx processing

DancingAxolotl:
## Related story
https://www.pivotaltracker.com/story/show/178148735

## Background
The initial sync of resolution service is very slow. Additionally, resolution service falls behind Polygon network if there are lots of events happening over a long period of time. 

## Changes
 - Added parallel processing of events for each domain

## Notes
Some potential improvements that can be made on top of this
 - Separate collection and processing of events (save event first, then process event)
 - Profiling/Optimization of DB queries
Also some additional testing would be good. I've only tested this on initial sync on Polygon starting from around block 22100000 where we had a spike of events.

### DIFF
--- a/src/models/CnsRegistryEvent.ts
+++ b/src/models/CnsRegistryEvent.ts
@@ -133,9 +133,11 @@ export default class CnsRegistryEvent extends Model {
 
   async beforeValidate(): Promise<void> {
     const tokenId = this.tokenId();
-    this.node = tokenId
-      ? CnsRegistryEvent.tokenIdToNode(BigNumber.from(tokenId))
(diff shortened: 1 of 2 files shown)


DancingAxolotl:
@sammyluo FV
This is an old optimisation idea that is still relevant for the service. Basically, it cuts down on initial sync and re-sync time by handling domain events in parallel. It will also help with processing huge spikes of domain events. For example, we had something like this happen during one of our holiday sales, the service was falling back from the blockchain since it couldn't process events fast enough.

This PR is a proof of concept and needs additional testing.

PR #66, opened 2022-02-04
Migration to set tlds domains parent_id as null

enaqx:
- Migration to set TLDs domains parent_id as null
- Update domains controllers tests accordingly
- Move env files under ignore

https://www.pivotaltracker.com/story/show/180830631


DeRain (approved):
Great job!

PR #72, opened 2022-02-23
Refactoring metadata response

refats:


refats:
https://www.pivotaltracker.com/n/projects/2465324/stories/180411972
refactored unit tests for new attributes
changed metaData response, deleted image_data field from OpenSeaMetadata
added basic attributed with domain ending and name length
also new attribute added for blockchain resolution if it's possible.

src/controllers/MetaDataController.ts
```diff
@@ -320,65 +305,25 @@ export class MetaDataController {
...
-    const attributes = [
-      ...this.getBasicDomainAttributes(name, meta),
-      ...this.getAnimalAttributes(name),
-    ];
-    if (attributes.find((attribute) => attribute.trait_type === 'animal')) {
-      domainType = 'animal';
-    }
-    if (PremiumDomains.includes(name)) {
-      domainType = 'premium';
-    }
```
DancingAxolotl, line 340: I think we should keep animal and premium attributes. Animal domains have custom domain images, so I think it would be appropriate to keep the domain type and animal type for these domains. Let's discuss it with the team later.

src/controllers/MetaDataController.ts
```diff
@@ -320,65 +305,25 @@ export class MetaDataController {
...
-    attributes.push({ trait_type: 'type', value: domainType });
+  private getDomainAttributes(domain: Domain): OpenSeaMetadataAttribute[] {
+    const attributes = this.getBasicDomainAttributes(domain.name);
+    const resolution = getDomainResolution(domain);
+    attributes.push({ trait_type: 'chain', value: resolution.blockchain });
```
DancingAxolotl, line 311: Maybe let's change `getBasicDomainAttributes(name: string)` to `getBasicDomainAttributes(domain: Domain)` and get the blockchain data there?

src/controllers/MetaDataController.ts
```diff
@@ -320,65 +305,25 @@ export class MetaDataController {
...
-      {
-        trait_type: 'level',
-        value: name.split('.').length,
+        trait_type: 'domain ending',
+        value: name.split('.')[1],
       },
       {
+        display_type: 'number',
         trait_type: 'length',
         value: punycode.toUnicode(name).split('.')[0].length,
```
DancingAxolotl, line 324: It's better to use `domain.extension` and `domain.label` instead of splitting the name manually

DancingAxolotl, line 324: It's better to use `domain.extension` and `domain.label` instead of splitting the name manually

DancingAxolotl (changes requested):
@refats Good work!
I've left a few minor comments. And I'd like to discuss the domain attributes for animal and premium domains. But other than that the PR looks good.

DancingAxolotl (changes requested):
@refats Good work!
I've left a few minor comments. And I'd like to discuss the domain attributes for animal and premium domains. But other than that the PR looks good.

src/controllers/MetaDataController.ts
refats: Yes, I can move part of getting resolution of domain here, if domain is valid.
But this need to change upper function defaultMetaResponse to provide Domain type instead the name.
No problem

src/controllers/MetaDataController.ts
refats: for example for test with unknown.crypto defaultMetaResponse couldn't create Domain object, so just parse the string name of provided input.

DancingAxolotl:
@sammyluo I think it should be okay to close this as stale.

PR #78, opened 2022-03-18
Add Binance, Avalanche, Fantom chains support

udhriaznov:
## Background

We want to support as many chains as possible and Moralis API gives us the ability to support not just Ethereum and Polygon networks, but several others as well. 

## Related Story / Issue

https://linear.app/unstoppable-domains/issue/ECUX-94/nft-api-add-support-for-all-moralis-chains

## Changes

- Added support in our NFT API for Binance, Avalanche and Fantom chain from the Moralis API

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -102,6 +102,15 @@ export const env = {
         process.env.POLYGON_ACCEPTABLE_DELAY_IN_BLOCKS || 100,
       ),
     } as EthUpdaterConfig,
+    BINANCE: {
+      JSON_RPC_API_URL: process.env.BINANCE_JSON_RPC_API_URL,
+    },
+    AVALANCHE: {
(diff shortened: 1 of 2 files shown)


iurevych (approved):
LGTM but let's also get @sudoryan review. I'm also wondering where do we get JSON RPC URLs from.

sudoryan:
LGTM! 

The JSON RPC URLS env config is stored under https://github.com/unstoppabledomains/resolution-service/settings/secrets/actions

udhriaznov:
@sudoryan They are EVM compatible. And, as I see in their Blockchain Explorers, they work with erc721.
Binance Smart Chain - https://bscscan.com/
Avalanche - https://snowtrace.io/
Fantom - https://ftmscan.com/

Binance Smart Chain also supports erc1155, but I'm not sure about other two chains.

sudoryan (approved):
LGTM

PR #79, opened 2022-03-25
[DT-676] Resolution service bridging hotfixes

DancingAxolotl:
## Related story
https://linear.app/unstoppable-domains/issue/DT-676/avatars-wont-show-for-bridged-domains

## Background
It seems like L2 bridge does not call `mint` to create tokens. It transfers the token directly from a null address, so the `NewURI` event is not emitted. This way L2 bridged domains are not recorded correctly in the resolution service.

## Changes
 - Added handling of Transfer events from Null addresses if the domain already exists as if it was a new URI
 - Added missing env variable for staging deployment

We will need to resync the service starting from domain bridging release to fix existing domains.


src/workers/eth/EthUpdater.ts
```diff
@@ -146,6 +146,20 @@ export class EthUpdater {
...
+      // domain exists, so it's probably a bridge
+      const resolution = domain.getResolution(this.blockchain, this.networkId);
+
+      resolution.ownerAddress = event.args?.to.toLowerCase();
+      resolution.registry = this.cnsRegistry.address;
+
+      const contractAddress = event.address.toLowerCase();
+      if (contractAddress === this.unsRegistry.address.toLowerCase()) {
+        resolution.resolver = contractAddress;
+        resolution.registry = this.unsRegistry.address.toLowerCase();
```
sudoryan, line 159: Maybe it was intentional for readability but this could be either:

```
resolution.resolver = this.unsRegistry.address.toLowerCase();
resolution.registry = this.unsRegistry.address.toLowerCase();
```

or 

```
resolution.resolver = contractAddress;
resolution.registry = contractAddress;

```

sudoryan (approved):
LGTM!

PR #88, opened 2022-04-01
[DT-656] Resolve by records

DancingAxolotl:
## Related Story
https://linear.app/unstoppable-domains/issue/DT-656/add-lookup-by-resolution-records-to-resolution-service

## Changes
 - Added `resolution` parameter to `/domains` endpoint
 - Made the owner parameter optional
 - Added `created_at` srting parameter
 - Restructured tests, split into multiple files.
   -- Note: `DomainsControllerGetDomain.test.ts` and `DomainsControllerGetDomainTransfers.test.ts` are just extracted from the old test file. Diffs show them as new files, but they are unchanged.


sudoryan:
If this purpose of this PR is to allow partners to show domains to users in their wallets is the owners param not enough?

src/controllers/dto/Domains.ts
```diff
@@ -96,13 +123,21 @@ export class DomainsListQuery {
 
   get sort() {
     return {
-      column: DomainsListQuery.SortFieldsMap[this.sortBy],
+      columns: DomainsListQuery.SortFieldsMap[this.sortBy].sortBy,
       direction: this.sortDirection,
     };
   }
 
+  nextStargingAfter(domain: Domain | undefined) {
```
sudoryan, line 131:
```suggestion
  nextStartingAfter(domain: Domain | undefined) {
```

src/controllers/DomainsController.ts
```diff
@@ -150,10 +175,11 @@ export class DomainsController {
         },
       });
     }
+
     response.meta = {
       perPage: query.perPage,
       nextStartingAfter:
-        lastDomain?.[query.sortBy]?.toString() || query.startingAfter || '',
+        query.nextStargingAfter(lastDomain) || query.startingAfter || '',
```
sudoryan, line 182:
```suggestion
        query.nextStartingAfter(lastDomain) || query.startingAfter || '',
```

src/controllers/DomainsController.ts
```diff
@@ -103,25 +101,52 @@ export class DomainsController {
       });
     }
 
+    if (query.resolution) {
+      const resolutionKeys = Object.keys(query.resolution);
+      for (let i = 0; i < resolutionKeys.length; i++) {
+        const key = Object.keys(query.resolution)[i];
+        where.push({
+          query: `"resolution"."resolution"->>'${key}' = :val${i}`,
```
sudoryan, line 109: Is this case sensitive?

sudoryan (approved):

DancingAxolotl:
> If this purpose of this PR is to allow partners to show domains to users in their wallets is the owners param not enough?

@sudoryan 
Yes, the original purpose is to show domains to users. But some partners want to look up domains by different records (e.g. btc, ltc addresses). This is not intended as a "reverse resolution" method, as anyone can add an address to their domain. It is purely for reference/visual display.
Besides showing user domains for different addresses, this can  be used to look up domains that resolve to a particular IPFS, dns, social, humanity check records etc.

src/controllers/DomainsController.ts
DancingAxolotl: Yes

PR #89, opened 2022-04-05
feat: export staging test domains script [DT-657]

enaqx:
Export test domains script from staging enviroment.


DancingAxolotl (approved):
LGTM

PR #92, opened 2022-04-06
Add OpenSea API support to fetch NFTs from eth chain

udhriaznov:
https://linear.app/unstoppable-domains/issue/ECUX-204/add-opensea-api-support-for-resolution-service-to-fetch-nfts-from-eth


src/env.ts
```diff
@@ -108,15 +108,6 @@ export const env = {
...
     } as EthUpdaterConfig,
-    BINANCE: {
-      JSON_RPC_API_URL: process.env.BINANCE_JSON_RPC_API_URL,
-    },
-    AVALANCHE: {
-      JSON_RPC_API_URL: process.env.AVALANCHE_JSON_RPC_API_URL,
-    },
-    FANTOM: {
-      JSON_RPC_API_URL: process.env.FANTOM_JSON_RPC_API_URL,
-    },
```
iurevych, line 119: Do we not need them anymore?

iurevych (approved):
👍

src/env.ts
udhriaznov: Yes, they were needed when we didn't have the Moralis API integration

PR #96, opened 2022-04-08
Check NFT ownership + use image_url instead of data

sudoryan:
2 Updates:

1. Checks if NFT PFP is actually owned by the domain's owner. This prevents domain owners from setting an NFT they don't own as their PFP.
2. Uses NFT image_url instead of image_data for background image. This could help with response times.


udhriaznov (approved):
LGTM

PR #102, opened 2022-04-11
Fix env var parsing

DancingAxolotl:
## Background

It seems like we have some problems with parsing env vars. Empty values are passed from deployment scripts if the value is not defined. In that case the resulting number is `NaN` which may cause issues for workers and the service in general.

For example, the staging matic worker is failing to sync because `POLYGON_BLOCK_FETCH_LIMIT` is not defined.

## Changes

This PR adds an additional `isNaN` check for numeric env variables to prevent invalid values from being saved to configs.


udhriaznov (approved):
Good catch! 👍

PR #104, opened 2022-04-11
Make the metadata api optional

DancingAxolotl:
## Background

Currently the `API` running mode launches both the metadata and resolution endpoints. Metadata endpoints are not necessary for most 3rd party providers. The metadata API also requires Moralis API keys which is an extra dependency.
We should make the metadata API optional using the running mode variables.

## Changes

- Added new running modes `SERVICE_API` and `METADATA_API` to run APIs separately.
- Moved check for Moralis API env variables to API setup.
- Updated the readme

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -14,11 +14,33 @@ import BugsnagPluginExpress from '@bugsnag/plugin-express';
 import { env } from './env';
 import ErrorHandler from './errors/ErrorHandler';
 
+const enabledControllers = [];
+
+if (
+  env.APPLICATION.RUNNING_MODE.includes('API') ||
+  env.APPLICATION.RUNNING_MODE.includes('SERVICE_API')
+) {
+  enabledControllers.push(DomainsController);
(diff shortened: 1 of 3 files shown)


src/api.ts
```diff
@@ -14,11 +14,33 @@ import BugsnagPluginExpress from '@bugsnag/plugin-express';
...
+) {
+  enabledControllers.push(DomainsController);
+  enabledControllers.push(StatusController);
+}
+
+if (
+  env.APPLICATION.RUNNING_MODE.includes('API') ||
+  env.APPLICATION.RUNNING_MODE.includes('METADATA_API')
+) {
+  if (!(env.MORALIS.API_URL && env.MORALIS.APP_ID)) {
```
udhriaznov, line 31: Could you please also add the same check for `OPENSEA_API_KEY` variable since we added the OpenSea API support recently?

udhriaznov (approved):
LGTM, thanks! ✨

PR #113, opened 2022-05-18
Fix .unstoppable fit

sudoryan:
Fixes .unstoppable TLD fit

# Before
Screen Shot 2022-05-18 at 2 34 32 PM

# After
Screen Shot 2022-05-18 at 2 35 02 PM


udhriaznov (approved):
Thanks for improving it 👍

PR #117, opened 2022-06-02
[DT-687] feat: Add support for reverse resolution

DancingAxolotl:
## Related story
https://linear.app/unstoppable-domains/issue/DT-687/support-reverse-resolution-in-resolution-service

## Changes
 - Added reverse resolution db model
 - Handled reverse resolution events in ETH worker
 - Added endpoints for reverse resolution
 - Upgraded uns lib version (contracts & sandbox)
 - Added migration to backfill reverse resolution records
 - Added tests


src/workers/eth/EthUpdater.ts
```diff
@@ -323,6 +328,53 @@ export class EthUpdater {
...
+
+  private async processRemoveReverse(
+    event: Event,
+    reverseRepository: Repository<DomainsReverseResolution>,
+  ): Promise<void> {
+    const args = unwrap(event.args);
+    const { addr } = args;
+
+    const reverseResolution = await DomainsReverseResolution.findOne({
+      where: { reverseAddress: addr.toLowerCase() },
```
sammyluo, line 368: We can set a transformer in the `reverseAddress` column in the `DomainsReverseResolution` table. Something like
```
  @Column('text', {transformer: lowercaseTransformer})
  reverseAddress: string;
```

sammyluo, line 368: Is it intentional that we don't need to specify the blockchain and network id here?

src/workers/eth/EthUpdater.ts
```diff
@@ -388,6 +440,17 @@ export class EthUpdater {
...
             break;
           }
+          case 'SetReverse': {
+            await this.processSetReverse(event, domainRepository);
+            break;
+          }
+          case 'RemoveReverse': {
+            await this.processRemoveReverse(
+              event,
+              manager.getRepository(DomainsReverseResolution),
```
sammyluo, line 450: nit: why not having a variable just like `domainRepository` to be consistent?

```suggestion
              const domainReverseResolutionRepository = manager.getRepository(Domain);
              
              this.processRemoveReverse(
                event,
                domainReverseResolutionRepository,
              )
```

sammyluo (changes requested):
Thanks for updating it! Exciting feature! 🙌 

It's a good start. Left some feedback.

src/workers/eth/EthUpdater.ts
DancingAxolotl: I don't have any preference, but my reasoning to have it inline was that it's more compact this way.
It doesn't make sense to define it at the beginning of the function like `domainRepository` since it's only used here. 
And if someone ever needs to reuse this repository, they can add a variable where necessary.

If you think it's better to have a variable declared, I'll update it.

sammyluo: Make sense. No, I don't have strong preference here either. Let's leave it as it.

src/database/migrations/1654097320150-BackfillReverseResolutions.ts
```diff
@@ -0,0 +1,59 @@
...
+    });
+    const reverseResolutions: Record<
+      string,
+      DomainsReverseResolution | undefined
+    > = {};
+    for (const event of events) {
+      switch (event.type) {
+        case 'SetReverse': {
+          const { addr, tokenId } = event.returnValues;
+          const domain = await Domain.findOne({ node: tokenId });
```
sammyluo, line 29: This is a N+1 query that could potentially affect the DB performance when this migration is running. We could optimize it so that once we get all the `events`, we query the list of corresponding domains right away, then processing them in-memory. Something like

```
const events = await queryRunner.manager.find(CnsRegistryEvent, {
  where: {
    type: ['SetReverse', 'RemoveReverse'],
  },
  order: {
    createdAt: 'ASC',
  },
});
const domainTokenIds = events
  .filter(({ type }) => type === 'SetReverse')
  .map(({ returnValues }) => returnValues.tokenId);

const domains = await Domain.find({ where: { node: In(domainTokenIds) } });
```

Then we can process and save the `DomainsReverseResolution` with `events` and `domains`

src/models/DomainsReverseResolution.ts
```diff
@@ -0,0 +1,32 @@
...
+import { Attributes, Blockchain } from '../types/common';
+import { ETHAddressRegex } from '../utils/ethersUtils';
+import { lowercaseTransformer } from '../database/transformers';
+
+@Entity({ name: 'domains_reverse_resolution' })
+@Unique(['domain', 'blockchain', 'networkId'])
+@Index(['domain', 'blockchain', 'networkId', 'reverseAddress'])
+export default class DomainsReverseResolution extends Model {
+  @Matches(ETHAddressRegex) // ensure that the address is lowercase to be consistent
+  @Column('text', { transformer: lowercaseTransformer })
```
sammyluo, line 13: Sorry I should've explained it more clearly in my previous comment, but what I was envisioning was to have comments in the DB layer, something like
```suggestion
  @Column('text', { transformer: lowercaseTransformer, comment: 'unique index, the ethereum address that is configured for reverse resolution' })
```
so it would be beneficial for someone who operates the DB without checking the source code.

src/models/Domain.test.ts
```diff
@@ -240,4 +242,147 @@ describe('Domain', () => {
       expect(fromDb?.parent?.name).to.equal('crypto');
     });
   });
+
+  describe('reverse resolution', () => {
```
sammyluo, line 246: Beautiful tests! 🙌

sammyluo (changes requested):

src/models/DomainsReverseResolution.ts
DancingAxolotl: Yes, This makes more sense. For some reason I assumed inline comments

PR #119, opened 2022-06-07
feat: new endpoint to list supported TLDs

sammyluo:
Create a new endpoint to list the supported Unstoppable Domains TLDs for resolution.


enaqx (approved):
LGTM 🎸 
Might be worth to add some test.

PR #125, opened 2022-07-06
Bulk domain records endpoint

iurevych:
This PR introduces a bulk domain records endpoint, as requested by the mobile team.

This endpoint can pull domain records for up to 50 domains in a bulk.
The endpoint also supports optional `key` query param to only pick certain keys out of they resolver keys we support.

Usage example:
`/records?domains[]=domain1.crypto&domains[]=domain2.crypto&key=social.picture.value`

Response where one domain has a picture record, and another domain does not even exist at UD:
records?domains =ryan crypto domains =iurevych crypto key=social picture value 2022-07-07 01-42-22

TODO:

- [x] Tests
- [x] Docs


src/controllers/MetaDataController.ts
```diff
@@ -263,6 +301,31 @@ export class MetaDataController {
     return metadata;
   }
 
+  @Get('/records')
+  @ResponseSchema(DomainsRecords)
+  @UseBefore(ConvertArrayQueryParams('domainNames'))
+  async getDomainsRecords(
+    @QueryParams() query: DomainsRecordsQuery,
+  ): Promise<DomainRecords[]> {
```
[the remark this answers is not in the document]
iurevych: Just did, please have a look.

DancingAxolotl (commented):
Some test for the new endpoint would be nice. Otherwise looks good!

src/services/Resolution.ts
```diff
@@ -27,7 +27,11 @@ export function getDomainResolution(domain: Domain): DomainsResolution {
     );
   }
 
-  if (isNullAddress(resolution.ownerAddress) && IsZilDomain(domain.name)) {
+  if (
+    !resolution.hasId() &&
+    isNullAddress(resolution.ownerAddress) &&
+    IsZilDomain(domain.name)
+  ) {
```
iurevych, line 34: @DancingAxolotl there is currently a bug where if you have two domains (one ETH and one ZIL) both with the same name, say `mike.zil`, we won't return ETH domain. As a fix I'm taking ETH zil domain as a priority.

Please let me know if this is correct logic.

iurevych:
@DancingAxolotl just reworked the implementation to follow new zil domains logic, please re-review.

src/services/Resolution.ts
mvlabat: @iurevych I believe `domain.name` includes its TLD

mvlabat: What this logic is missing though is `.zil` support on Ethereum.

iurevych: @mvlabat during my tests I saw this logic working perfectly fine. Could you clarify what exactly is wrong with the code?

The code currently first attempts to fetch MATIC domain resolution.
Then it takes ETH domain resolution.
Finally, when nothing found, and the domain name ends with `.zil`, we attempt to fetch ZIL domain resolution

DancingAxolotl: I think this won't solve the bug. As far as I understand, the problem is the following:
We may have two domains:
```
domain1 = {
name: 'mike.zil'
node: '0x0123' //zns-namehash
resolutions: [{blockchain: ZIL, ...}]
}
domain2 = {
name: 'mike.zil'
node: '0x0567' //uns-namehash
resolutions: [{blockchain: MATIC, ...},{blockchain: ETH, ...}]
}
```
In which case `Domains.findOne('mike.zil")` will return a random domain (probably most recent one, with the highest I'd). And while `getDomainResolutions` may work correctly, it won't see one set of resolutions. So it will either return only ZIL or only ETH.

I guess this will be a problem if somehow a zil domain is bridged to UNS while it is not burned in ZNS. But the fix should be different.

DancingAxolotl: The proper fix would be to either lookup the domain by namehash, group domains by name instead of using `findOne` or change the model to have multiple `node` values.

DancingAxolotl: @iurevych do you think this bug will affect any domains in production? If the impact is not big, i think you can just create a ticket for dev tools and we will fix the logic later.

iurevych: @DancingAxolotl thanks Mike, I reverted it since this was something I personally noticed when doing local tests, filed an issue https://linear.app/unstoppable-domains/issue/DT-820/support-zil-domains-minted-on-both-zil-and-matic

src/controllers/DomainsController.ts
```diff
@@ -10,17 +10,22 @@ import {
...
   DomainsListResponse,
   UnsDomainQuery,
   DomainLatestTransferResponse,
+  DomainsRecordsQuery,
+  DomainsRecordsResponse,
+  DomainRecords,
 } from './dto/Domains';
 import { ConvertArrayQueryParams } from '../middleware/ConvertArrayQueryParams';
 import { In } from 'typeorm';
+import _ from 'lodash';
```
sammyluo, line 27:
```suggestion
import pick from 'lodash/pick';
```

sammyluo (approved):
LGTM! Just some minor suggestions.

PR #128, opened 2022-07-08
Implement test upload route for PFP NFT

markvital:
Created test API endpoint for bechmarking PFP NFT uploading to CDN in production.

TODO:

- [x] test on staging to check GCS config is working
- [x] write docs how to run GSC emulator locally
- [x] deploy to prod, to do benchmarking

## Related task:
https://linear.app/unstoppable-domains/issue/ECUX-297/implement-test-upload-route-for-pfp-nft


markvital:
Tested on staging, it seems to be working...
[https://resolve.staging.unstoppabledomains.com/test_upload/markvital.x](https://resolve.staging.unstoppabledomains.com/test_upload/markvital.x)

First request uploads image to Google storage:

Screen Shot 2022-07-09 at 12 48 14 AM

Second request gets image from our CDN:

Screen Shot 2022-07-09 at 12 49 30 AM

src/controllers/UploadTestController.ts
```diff
@@ -0,0 +1,104 @@
...
+  @IsNotEmpty()
+  @IsString()
+  domain: string;
+
+  @IsString()
+  pfpImage?: string;
+
+  @IsNotEmpty()
+  @IsString()
+  status: string;
```
iurevych, line 30: I'd suggest using enum instead of a string, just to be safe. Example:
```ts
enum UploadStatus {
  Found = 'FOUND',
  Cached = 'FOUND_IN_CACHE',
  NotFound = 'NOT_FOUND'
}

class UploadResponse {
  @IsNotEmpty()
  @IsString()
  status: UploadStatus;
}
```

src/controllers/UploadTestController.ts
```diff
@@ -0,0 +1,104 @@
...
+
+      if (resolution && socialPic) {
+        const { contractAddress, tokenId } = parsePictureRecord(socialPic);
+        const nftPfpFolder = 'nft-pfp';
+        const fileName = `${nftPfpFolder}/${contractAddress}_${tokenId}`;
+        const bucketName = env.CLOUD_STORAGE.CLIENT_ASSETS.BUCKET_ID;
+        const hostname =
+          env.CLOUD_STORAGE.API_ENDPOINT_URL ||
+          'https://storage.googleapis.com';
+        const imageURL = `${hostname}/${bucketName}/${fileName}`;
```
iurevych, line 57: Does it work without a file extension?

iurevych (approved):
👍 lovely, let's give it a try! :shipit:

src/controllers/UploadTestController.ts
markvital: Yes, it does. Apparently OpenSeas doesn't always use extension to cache in their CDN..
I guess binary file stores mime type metadata inside.
Otherwise I can make a folder with this name, and put a file with extension, but this would require more sophisticated query: checking if folder exists, than reading its content (ls) - not optimal

PR #129, opened 2022-07-13
[DT-746] fix: ZIL resolution on UNS

DancingAxolotl:
## Changes
 - Allowed duplicated domain names, only `node` (namehash) must be unique.
 - Updated resolution logic in `controllers/MetaDataController` and `services/Resolution`
 - Updated/added tests

### DIFF
--- a/src/controllers/DomainsControllerGetDomain.test.ts
+++ b/src/controllers/DomainsControllerGetDomain.test.ts
@@ -268,6 +268,35 @@ describe('DomainsController', () => {
       });
     });
 
+    it('should return zil domain on uns', async () => {
+      await DomainTestHelper.createTestDomain({
+        blockchain: Blockchain.ETH,
+        networkId: env.APPLICATION.ETHEREUM.NETWORK_ID,
+        name: 'sometestforzil.zil',
+        ownerAddress: '0xcea21f5a6afc11b3a4ef82e986d63b8b050b6910',
+        resolver: '0x34bbdee3404138430c76c2d1b2d4a2d223a896df',
+        registry: '0x9611c53be6d1b32058b2747bdececed7e1216793',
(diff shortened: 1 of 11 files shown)


src/controllers/MetaDataController.ts
```diff
@@ -489,22 +480,27 @@ export class MetaDataController {
...
 
-  private normalizeDomain(domain: string): string {
-    domain = domain.trim().toLowerCase();
+    let domain =
+      (await Domain.findByNode(tokenName)) ||
+      (await Domain.findOnChainNoSafe(tokenName));
 
-    if (domain.endsWith('.zil')) {
-      return znsNamehash(domain);
+    if (!domain && domainName.endsWith('.zil')) {
```
hivtushok, line 498: Are we sure that this relation on `.zil` extension is safe enough?

src/services/Resolution.ts
```diff
@@ -8,27 +8,29 @@ export function IsZilDomain(name: string): boolean {
...
     resolution = domain.getResolution(
-      Blockchain.ZIL,
-      env.APPLICATION.ZILLIQA.NETWORK_ID,
+      Blockchain.ETH,
+      env.APPLICATION.ETHEREUM.NETWORK_ID,
     );
-  } else {
+  }
+
+  if (isNullAddress(resolution.ownerAddress) && IsZilDomain(domain.name)) {
```
hivtushok, line 29: Same here – it would be great if we could avoid relying on the tlds.

src/controllers/DomainsControllerGetDomain.test.ts
```diff
@@ -268,6 +268,35 @@ describe('DomainsController', () => {
       });
     });
 
+    it('should return zil domain on uns', async () => {
+      await DomainTestHelper.createTestDomain({
+        blockchain: Blockchain.ETH,
```
hivtushok, line 273: Would be great if we could add a test for `Matic` as well 👍🏼

PR #132, opened 2022-07-26
Fix token ID matching

iurevych:
Per a recent PR review there was a suggestion to add a check for `startsWith`, assuming we're matching ETH addresses.

The issue here is that we stopped matching token IDs. So we're returning back to a previous lose matching:
```js
/^[a-fA-F0-9]+$/.test('6304531997610998161237844647282663196661123000121147597890468333969432655810'); // true
```


sammyluo (approved):
LGTM!

PR #136, opened 2022-08-26
Fixed nodemon dependency for debug conf. introducted in PR#124 and PR…

markvital:
Missed dev dependency for nodemon in debug conf.

I was getting some non-verbose error during launch:
mystery debug error

Now `start:dev:debug` and `test:debug` should work as expected


DancingAxolotl (approved):
LGTM

PR #137, opened 2022-08-27
Add avatars NFT PFP conversion from SVG to JPEG

markvital:
Implement SVG to PNG conversion using [convert-svg-to-jpeg](https://www.npmjs.com/package/convert-svg-to-jpeg) / [Puppeteer](https://github.com/puppeteer/puppeteer). SVG images would be used for twitter and open graph metadata for domain profiles.

After domain record social.picture.value is updated, we'll save 4 images in CDN:

- with overlay, SVG
- without overlay, SVG
- with overlay, JPG
- without overlay, JPG


src/utils/socialPicture/index.ts
```diff
@@ -148,14 +149,16 @@ export const cacheSocialPictureInCDN = async (
   }
   const fileName = getNFTFilenameInCDN(socialPic);
   const fileNameWithOverlay = getNFTFilenameInCDN(socialPic, domain.name);
+  const fileNameJpeg = getNFTFilenameInCDN(socialPic)?.replace('.svg', '.jpg'); // dirty hack, pass param instead
+  const fileNameWithOverlayJpeg = getNFTFilenameInCDN(socialPic, domain.name)?.replace('.svg', '.jpg');
```
iurevych, line 153: Consider using `getNFTPictureFname`, it already does the replacement.

src/utils/socialPicture/index.ts
```diff
@@ -148,14 +149,16 @@ export const cacheSocialPictureInCDN = async (
...
   const bucket = storage.bucket(bucketName);
 
   const [fileExists] = await bucket.file(fileName).exists();
-  const [fileWithOverlayExists] = await bucket
-    .file(fileNameWithOverlay)
-    .exists();
-  if (!fileExists || !fileWithOverlayExists) {
+  const [fileWithOverlayExists] = await bucket.file(fileNameWithOverlay).exists();
+  const [fileJpegExists] = await bucket.file(fileNameJpeg).exists();
+  const [fileWithOverlayJpegExists] = await bucket.file(fileNameWithOverlayJpeg).exists();
```
iurevych, line 160: Consider parallelising requests:
```suggestion
  const [fileExists , fileWithOverlayExists, fileJpegExists, fileWithOverlayJpegExists] = await Promise.all([
    bucket.file(fileName).exists(),
    bucket.file(fileNameWithOverlay).exists(),
    bucket.file(fileNameJpeg).exists(),
    bucket.file(fileNameWithOverlayJpeg).exists()
  ]);
```

src/utils/socialPicture/index.ts
```diff
@@ -185,10 +188,31 @@ export const cacheSocialPictureInCDN = async (
...
+          fetchedMetadata?.background_color || '',
+          true,
+        );
+        const withOverlayImageDataJpeg = await convert(withOverlayImageData); //TODO: increase resolution
+        files.push({ fname: fileNameWithOverlayJpeg, data: withOverlayImageDataJpeg });
+      }
+
       // @TODO: this actually doesn't wait for uploading to finish. Re-do so we wait
       // temporarity replaced Promise.all() with seqential execution instead of parallel
       for (const file of files) {
```
iurevych, line 214: Can we also run these requests in parallel?

src/utils/socialPicture/index.ts
iurevych: Alternatively, have you tried this?
```ts
const files = await bucket.getFiles({
  prefix: fileName
});
```

You can extend `getNFTFilenameInCDN` to support an argument to return a file name without an extension.

iurevych (changes requested):
Looks like this is what we'd want to test. Please see comments, we have to ensure requests are handled quickly.

src/controllers/MetaDataController.ts
```diff
@@ -645,26 +660,63 @@ export async function fetchTokenMetadata(
...
+    socialPicture, 
+    domain,
+    resolution,
+    withOverlay,
+    rasterize,
+    getNftPfpImageFromCDN
+  )
+}
+
+async function getOrCacheNowPfpNFTPath(
```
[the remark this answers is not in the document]
markvital: Because we need path to return path to redirect to the entry in the bucket and we need to return the binary of image (or string in case of SVG) in _/image/domain.x_ endpoint.
I actually didn't implement second use case properly yet. And we also have it in /metadata endpoint, where we embed the image in response (for the backward compatibility?)

src/utils/socialPicture/index.ts
markvital: Let me try that..

src/utils/socialPicture/index.ts
markvital: Yes, of course

src/utils/socialPicture/index.ts
markvital: Good eye, thank you

Dockerfile
```diff
@@ -20,5 +20,6 @@ RUN yarn cache clean
 
 # Cleanup development packages
 RUN yarn install --prefer-offline --frozen-lockfile --production --unsafe-perm --ignore-optional
+RUN yarn add @resvg/resvg-js-linux-arm64-musl --prefer-offline --frozen-lockfile --production --unsafe-perm --ignore-optional
```
mvlabat, line 23: I'm afraid this won't work for environments running on the arm64 architecture.

We should just remove the `--ignore-optional` flag from the command above.

src/controllers/MetaDataController.ts
```diff
@@ -281,6 +284,7 @@ export class MetaDataController {
           domain,
           resolution,
           withOverlay,
+          rasterize
```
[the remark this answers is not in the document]
markvital: Reworked by introducing `PfpNftParams`
```javascript
async function getOrCacheNowPfpNFT(nftParams: PfpNftParams) {
  return getOrCacheImage(nftParams, getNftPfpImageFromCDN);
}
```

Dockerfile
```diff
@@ -11,7 +15,10 @@ RUN yarn install --production=false --no-lockfile
 ADD . ./
 RUN yarn build
 
+# added this to fix error: Incorrect hash when fetching from the cache for "wyvern-js"
+RUN yarn cache clean
```
mvlabat, line 19: this shouldn't be merged to master. Afaik, this error is present only in your local environment, but not in CI

markvital:
Continuing work on this in [PR#172](https://github.com/unstoppabledomains/resolution-service/pull/172)

PR #138, opened 2022-09-07
[DT-801] docs: Updating README and env sample files

crazeekennee:
**Background**

The resolution-service's README.md file has gotten out of date, and needs some maintenance to be current with our resolution-service

**Changes**

We updated the README.md file to reflect the currently required environment variables, recommended service versions, service descriptions, and removed some ambiguity in the service description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ variety of platforms and in the cloud.
   RDS, Google Cloud SQL). To install postgres locally, follow these
   [instructions](https://www.postgresql.org/download). Make sure to configure
   password authentication for the DB user that will be used by the service.
+  - This guide has been tested with Postgres 11.17, although other versions may work as well.
 
 ### Quick start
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -55,25 +56,27 @@ variety of platforms and in the cloud.
 
 ```
 NODE_ENV=production
-RESOLUTION_POSTGRES_HOST=example.com:5432      # DB host
-RESOLUTION_POSTGRES_USERNAME=example           # DB user configured in postgres
-RESOLUTION_POSTGRES_PASSWORD=password          # DB password configured in postgres
+RESOLUTION_POSTGRES_HOST=example.com          	# DB host
+RESOLUTION_POSTGRES_PORT=5432                   # DB port
```
crazeekennee, line 60: Specifying RESOLUTION_POSTGRES_PORT separately from the host line,, to match convention with line 99.

README.md
```diff
@@ -209,6 +212,7 @@ Additional pre-requirements that are necessary for development:
 1. Install project dependencies
 
 ```
+nvm install 14.16.1
```
crazeekennee, line 215: This step was required for me, but I only knew about it after running `nvm use`, so why not specify it before, so it'll be installed before `nvm use` tries to use it ?

README.md
```diff
@@ -225,6 +229,10 @@ RESOLUTION_POSTGRES_DATABASE=resolution_service
 ETHEREUM_JSON_RPC_API_URL=localhost:8545
 POLYGON_JSON_RPC_API_URL=localhost:8546
 VIEWBLOCK_API_KEY=apikey
+METADATA_API=apikey
```
crazeekennee, line 232: Same comment as on line 67, application wouldn't start unless these value existed in local.dev.env

sammyluo (approved):
Looks great! Thanks for updating it! 🚀 ⭐

PR #139, opened 2022-09-09
Don't ignore optional dependencies in Dockerfile

mvlabat:
Ignoring optional dependencies causes the build process to avoid installing platform-specific dependencies.

This should fix the resvg lib integration.


markvital (approved):
All good, let's try to deploy it to staging!

PR #143, opened 2022-09-10
Cleanup and add attributes

sudoryan:
PR spawned from this slack thread: https://unstoppabledomains.slack.com/archives/CJ0LDPWS2/p1662521836037879

Attributes are meant to be a useful way to sort and filter through collections on Opensea. Many of our attributes are not at all useful and have unique values which act more like a metric than filter. This sort of clutter make our collection look messy and disorganized. 
Cleaning up our attributes is an easy win for our users. We can help the domain secondary market giving visibility and attention to hyped categories like 999, 10k, 100k clubs. With these new attributes they can sort these categories by price letting them see the floor and thus driving up hype and sales

These changes should not affect any of our partners or resolution as the attribute field is intended to be solely used for filtering on Opensea or any other secondary market platforms

## Changes

- Remove **domain** attribute
- Remove "invalid" property
- Capitalize attributes (looks a lot better)
- Adds **ending** attribute
- Adds **character set** attribute (digit, alphanumeric, letter)
- Adds 999 club, 10k club, 100k club domain types

This is what the attributes currently look like in Opensea:
Screen Shot 2022-09-10 at 10 13 03 AM


src/utils/metadata/index.ts
```diff
@@ -0,0 +1,103 @@
...
+  Digit = 'digit',
+}
+
+export enum DomainType {
+  Standard = 'standard',
+  Animal = 'animal',
+  Premium = 'premium',
+}
+
+export enum DomainCategory {
```
sammyluo, line 38: Would be great to rename it so that it's more specific. Otherwise `DomainCategory` may be confusing in other context.

src/utils/metadata/index.ts
```diff
@@ -0,0 +1,103 @@
...
+  '999Club' = '999 Club',
+  '10kClub' = '10k Club',
+  '100kClub' = '100k Club',
+}
+
+export const getNumberClub = (domain: Domain): DomainCategory | '' => {
+  const label = domain.name.split('.').shift();
```
sammyluo, line 45: Would be great to have it as a util function in src/utils/domain.ts. Can be reused in `getDomainCharacterSet` and it would be useful in the future as well

src/utils/metadata/index.ts
```diff
@@ -0,0 +1,103 @@
...
+  Premium = 'premium',
+}
+
+export enum DomainCategory {
+  '999Club' = '999 Club',
+  '10kClub' = '10k Club',
+  '100kClub' = '100k Club',
+}
+
+export const getNumberClub = (domain: Domain): DomainCategory | '' => {
```
sammyluo, line 44: For the return type, why not use `DomainCategory | null`? I find it easier to process in validations than string type. WDYT?

src/utils/domain.ts
```diff
@@ -6,10 +6,14 @@ const normalizeToken = (token: string): string => {
   return '0x' + BigInt(token).toString(16).padStart(64, '0');
 };
 
-export const normalizeDomainName = (domainName: string) => {
+export const normalizeDomainName = (domainName: string): string => {
   return domainName.trim().toLowerCase();
 };
 
+export const getDomainNameTld = (domainName: string): string => {
+  return domainName.split('.').pop() ?? '';
```
sammyluo, line 14: We might want to call `normalizeDomainName()` first.

src/utils/AnimalDomainHelper/AnimalDomainHelper.ts
```diff
@@ -32,13 +36,17 @@ export default class AnimalDomainHelper {
...
-      attributes.push({ trait_type: 'adjective', value: prefix });
+      attributes.push({ trait_type: DomainTraitType.Adjective, value: prefix });
     }
     if (animal) {
-      attributes.push({ trait_type: 'animal', value: animal });
+      attributes.push({ trait_type: DomainTraitType.Animal, value: animal });
     }
     return attributes;
   }
+  isAnimalDomain(domainName: string): boolean {
```
sammyluo, line 46: Nitpick: would be great to have a newline in front of the function for better readability.
```suggestion

  isAnimalDomain(domainName: string): boolean {
```

src/utils/metadata/index.ts
```diff
@@ -0,0 +1,103 @@
+import premiumDomains from './premium.json';
+import customImageDomains from './custom-image.json';
+import AnimalDomainHelper from '../AnimalDomainHelper/AnimalDomainHelper';
+import { Domain } from '../../models';
+
+const PremiumDomains: string[] = premiumDomains;
+const CustomImageDomains: Record<string, string> = customImageDomains;
+
+export enum DomainTraitType {
+  Adjective = 'Adjective',
```
sammyluo, line 10: I'm not very familiar with the Opensea metadata requirements, but are the values of these enums case-sensitive? Asking because it seems like these values were lower case previously.

sammyluo (changes requested):
Nice! Just left some minor change requests and questions. Otherwise LGTM!

src/utils/metadata/index.ts
sudoryan: It's not case sensitive. Changed to uppercase to make it look better

PR #151, opened 2022-09-16
Sre 648 resolve fastly conflicts with gae routes

UDtorrey:


UDtorrey:
merging as this is a fix to the GH Action workflow for staging deployment; currently getting "No space left on device" from GH action runners

PR #153, opened 2022-09-27
[DT-887] fix: reverse resolution records saving

DancingAxolotl:
## Background
Resolution service returns incorrect reverse resolutions when the reverse address is set multiple times. This happens because the old reverse resolution is not removed from db when we add a new one.

### Linear issue
https://linear.app/unstoppable-domains/issue/DT-887/resolution-service-returns-incorrect-reverse-resolutions

## Changes
 - Update `processSetReverse` in `EthUpdater`
 - Update tests


sammyluo (approved):
🙌

src/database/migrations/1664366869085-CleanReverseResolutions.ts
```diff
@@ -0,0 +1,23 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class CleanReverseResolutions1664366869085
+  implements MigrationInterface
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
```
enaqx, line 6: If this migration does not change db schema and is non-reversible mayve it is better to run directly on DB without migration?

DancingAxolotl: Makes sense. I'll add it to `scripts` instead to preserve the query. Thanks!

enaqx (approved):
LGTM! 👍

PR #155, opened 2022-09-29
feat: Add boolean "reverse" field to domain response

sammyluo:
Added a `reverse` field to domain json responses in APIs:
- GET /domains
- GET /domains/:domainName 
- GET /reverse/:address


enaqx (approved):
LGTM!

PR #157, opened 2022-10-03
Return based64 encoded SVG image for animal domain profiles

markvital:
in the /metadata endpoint we currently render animal image data starting with the wrapping SVG tag https://metadata.unstoppabledomains.com/metadata/nycdragon22.crypto
however all other pictures are rendered with correct base64 data https://metadata.unstoppabledomains.com/metadata/matt.crypto

nycdragon22 crypto 2022-10-03 16-27-28


iurevych (approved):
👍 let's please also fix the test:
resolution-service@3747739 2022-10-04 14-15-09

PR #162, opened 2022-10-11
chore: add TypeORM cache for findDomainByNameOrToken()

sammyluo:
Add cache to reduce DB queries generated from the /metadata endpoints


src/utils/domain.ts
```diff
@@ -37,11 +37,11 @@ export const findDomainByNameOrToken = async (
   const domainName = normalizeDomainName(domainOrToken);
 
   let domain =
-    (await Domain.findByNode(tokenName)) ||
+    (await Domain.findByNode(tokenName, undefined, true)) ||
     (await Domain.findOnChainNoSafe(tokenName));
```
DancingAxolotl, line 41: @sammyluo what do you think about removing this lookup. It performs a Blockchain query in case the domain is not picked up by the service yet. Since we reduced block delays for the service it shouldn't be a problem anymore. The external lookup probably adds more latency than the DB query in this case.

DancingAxolotl (approved):
LGTM, just added one more suggestion for performance improvement which we can consider. But we can do that in a separate PR

PR #169, opened 2022-10-21
Boost NTF PFP image fetching timeout

markvital:
As per the earlier discussion we decided to boost the PFP NFP image timeout to 30 sec. This will allow loading bigger images, especially the one hosted on IPFS.
Since the fetching would be mostly done by the worker and is done only once, this should not impact overall performance.


src/utils/socialPicture/index.ts
```diff
@@ -60,7 +60,10 @@ export const getNFTSocialPicture = async (
     return [base64, mimeType];
   }
 
-  const resp = await nodeFetch(makeImageLink(pictureOrUrl), { timeout: 5000 });
+  const FETCHING_TIMEOUT = 30 * 1000; // timeout in ms
+  const resp = await nodeFetch(makeImageLink(pictureOrUrl), {
+    timeout: FETCHING_TIMEOUT,
+  });
```
DancingAxolotl, line 66: What do you think about adding this into env configs, so we can tweak it without changing the code?

Did we backfill PFP images for existing tokens? If not, this may cause significant latency for `/image` and `/image-src` endpoints for tokens which were not cached yet since `getNFTSocialPicture` may be called there too.

iurevych:
Closing this PR until someone on the team will take it.

PR #170, opened 2022-10-21
fix: case insensitive resolution search for /domains endpoint

sammyluo:
Case insensitive resolution search for /domains endpoint


src/controllers/DomainsController.ts
```diff
@@ -137,7 +137,7 @@ export class DomainsController {
       for (let i = 0; i < resolutionKeys.length; i++) {
         const key = Object.keys(query.resolution)[i];
         where.push({
-          query: `"resolution"."resolution"@>'{"${key}":"${query.resolution[key]}"}'::jsonb`,
+          query: `lower(("resolution"."resolution")::text)::jsonb @> lower('{"${key}":"${query.resolution[key]}"}')::jsonb`,
```
DancingAxolotl, line 140: We may want to double-check that this doesn't cause a full table scan on the resolution table.
You can probably check it by running the query manually with `explain` locally or on the staging DB.

DancingAxolotl (approved):
LGTM!
I have one concern regarding how this change may affect performance. It may be worth double checking, but I think it should be okay.

src/controllers/DomainsController.ts
sammyluo: Good callout! Here's the execution plan for this query:

```
Nested Loop Left Join  (cost=8.93..18.63 rows=1 width=706)
  Join Filter: ("reverseResolutions".domain_id = domain.id)
  ->  Nested Loop  (cost=8.93..17.49 rows=1 width=614)
        ->  Merge Right Join  (cost=8.64..9.17 rows=1 width=222)
              Merge Cond: (parent.id = domain.parent_id)
              ->  Index Scan using "PK_05a6b087662191c2ea7f7ddfc4d" on domains parent  (cost=0.28..260.53 rows=5415 width=111)
              ->  Sort  (cost=8.31..8.32 rows=1 width=111)
                    Sort Key: domain.parent_id
                    ->  Index Scan using "PK_05a6b087662191c2ea7f7ddfc4d" on domains domain  (cost=0.28..8.30 rows=1 width=111)
                          Index Cond: (id = 11)
        ->  Index Scan using "UQ_16efa381afd8187533c52240211" on domains_resolution resolution  (cost=0.28..8.32 rows=1 width=392)
              Index Cond: (domain_id = 11)
              Filter: ((lower((resolution)::text))::jsonb @> '{"crypto.eth.address": "0xc2cc046e7f4f7a3e9715a853fc54907c12364b6b"}'::jsonb)
  ->  Seq Scan on domains_reverse_resolution "reverseResolutions"  (cost=0.00..1.12 rows=1 width=92)
        Filter: (domain_id = 11)
```

It's still using an index scan for domains_resolution.resolution. The lower and casting only applies to the filter.

Will still keep an eye on the performance 👍

PR #171, opened 2022-10-21
[DT-917] feat: add api key enrolment to resolution

DancingAxolotl:
## Background
Currently there is no way to add API keys besides manually changing the DB. This PR introduces an endpoint to add new keys to the service.

## Related story
https://linear.app/unstoppable-domains/issue/DT-917/add-api-key-enrolment-to-resolution-service

## Changes
 - Added new endpoint
 - Changed apiKey db type to `test` instead of `uuid` to support external keys
 - Added tests


SteversIO:
Looks like tests are failing.

.github/workflows/tests.yml
```diff
@@ -45,6 +45,7 @@ jobs:
       MORALIS_APP_ID: ${{ secrets.MORALIS_APP_ID }}
       OPENSEA_API_KEY: ${{ secrets.OPENSEA_API_KEY}}
       CLOUD_STORAGE_BUCKET_ID: 'resolution-client-assets'
+      RESELLER_APP_KEY: 'test-app-key'
```
SteversIO, line 48: learning question: Why do we have OpenSea and Moralis API keys coming from config but this is hard coded? Are we mixing unit and integration tests?

src/controllers/EnrolmentController.test.ts
```diff
@@ -0,0 +1,112 @@
+import supertest from 'supertest';
+import { api } from '../api';
+import { expect } from 'chai';
+import { env } from '../env';
+import ApiKey from '../models/ApiKey';
+
+describe('EnrolmentController', () => {
```
SteversIO, line 7: English word is Enroll. Is there a different meaning for Enrol?  Remind me to share the broken window theory on next standup

src/controllers/EnrolmentController.test.ts
```diff
@@ -0,0 +1,112 @@
...
+          expectedError: { isNotEmpty: 'name should not be empty' },
+        },
+        {
+          desc: 'should validate non-string name',
+          body: { key: testKey, name: 11 },
+          expectedError: { isString: 'name must be a string' },
+        },
+      ];
+
+      cases.forEach((c) => {
```
SteversIO, line 92: Do we have a style guide for JS/TS?

I would expect a variable name like `c` to violate it.

src/controllers/EnrolmentController.test.ts
```diff
@@ -0,0 +1,112 @@
...
+import { expect } from 'chai';
+import { env } from '../env';
+import ApiKey from '../models/ApiKey';
+
+describe('EnrolmentController', () => {
+  describe('POST /enrol', () => {
+    const testKeyName = 'testname';
+    const testKey = 'testkey';
+
+    it('should check auth', async () => {
```
SteversIO, line 12: Great structure to tests 💪🥳

SteversIO (changes requested):
Some questions, some feedback. Happy to move my general questions to team chat.

src/controllers/EnrolmentController.test.ts
DancingAxolotl: Apparently my spell checker is British :D
https://grammarist.com/spelling/enrol-enroll/

src/controllers/EnrolmentController.test.ts
DancingAxolotl: I'm not sure. Even if there is one, I don't think it's enforced particularly well...

I backed myself into a corner with this one. The array is named `cases` so it would be appropriate to name the iterator `case`, but that's a reserved word so it ended up being just `c` :D 
`testCases` and `testCase` should be more appropriate, thanks for catching!

.github/workflows/tests.yml
DancingAxolotl: Great question! As far as I can tell, we don't actually call external APIs in unit tests. In fact, it looks like image caching is not covered with unit tests at all :D 

I think it should be safe to replace these with dummy keys. And the pfp stuff should get covered by unit tests eventually as we will be pushing to improve test coverage for the service soon.

SteversIO (approved):
Conditionally approving under the pretense that I don't fully understand the role that `up` and `down` play in the `database/migrations` folder. But the queries seem straight forward.  I saw you're changing the apiKey from a UUID to a text.

src/controllers/EnrollmentController.ts
```diff
@@ -0,0 +1,43 @@
...
+import { getConnection } from 'typeorm';
+import { APIKeyEnrollmentParams } from './dto/Enrollment';
+import { env } from '../env';
+
+@JsonController()
+export class EnrollmentController {
+  @Post('/enroll')
+  async postEnroll(
+    @Body() params: APIKeyEnrollmentParams,
+    @HeaderParam('reseller-app-token') token: string,
```
sammyluo, line 20: Let's rename the header to `app-auth-token` and change the related variables and env variables.

sammyluo (changes requested):
One small change request. Otherwise LGTM! 🚀 ⭐

enaqx (approved):

sammyluo (approved):

PR #172, opened 2022-10-21
Convert NFT PFT SVG to JPG using Puppeteer / Browserless

markvital:
In [PR#137](https://github.com/unstoppabledomains/resolution-service/pull/137) conversion of PFP avatars from SVG to JPG was implemented using 9convert-svg-to-jpeg](https://www.npmjs.com/package/convert-svg-to-jpeg) which is a very simple library, that startup Puppeteer locally.
Unfortunately, we were unable to start Puppeteer on GCS Compute Engine easily and decided to go with the SAAS solution instead.
Decided to go with [Browserless](https://www.browserless.io), which provides a robust production-ready wrapper around the Puppeteer providing a pool of warmed-up headless browsers instances and has a SAAS solution.

Here is how the conversion works locally. See [README](https://github.com/unstoppabledomains/resolution-service#readme) for details.

Screenshot 2022-10-21 at 20 36 03


src/controllers/MetaDataController.ts
```diff
@@ -273,15 +277,16 @@ export class MetaDataController {
...
     if (domain && resolution) {
-      const socialPictureValue = resolution.resolution['social.picture.value'];
+      const socialPicture = resolution.resolution['social.picture.value'];
       const pfpImageFromCDN =
-        socialPictureValue &&
-        (await getOrCacheNowPfpNFT(
-          socialPictureValue,
+        socialPicture && (
+          await getOrCacheNowPfpNFT({
+          socialPicture,
```
DancingAxolotl, line 284: In the `/metadata/:domainOrToken` endpoint it looks like we use `getNftPfpImageFromCDN` instead of `getOrCacheNowPfpNFT` is there any reason to use one over the other? Should we use `getOrCacheNowPfpNFT` in the `/metadata` endpoint too?

Also, do you think we could remove the `image_data` field from the `/metadata` response entirely? If it has to generate an image it may cause high latency of the endpoint which may affect performance of the API. Since we already have the `image_url` there, is there a need to keep the image data in the response?

src/utils/socialPicture/index.ts
```diff
@@ -185,43 +212,119 @@ export const cacheSocialPictureInCDN = async (
...
+          fname: fileNameWithOverlayJpeg,
+          data: withOverlayImageData,
+          shouldConvert: true,
+        });
       }
+
+      await Promise.all(
+        files.map(async ({ fname, data, shouldConvert }) => {
+          if (shouldConvert) {
+            const puppeteerURL = env.PUPPETEER.WS_URL || 'ws://localhost:3000';
```
DancingAxolotl, line 246: Should we check if puppeteer is running during service startup?
Is it possible to make jpeg conversion optional?

local.dev.env.sample
```diff
@@ -15,4 +15,10 @@ MORALIS_API_URL=http://localhost
 MORALIS_APP_ID=1
 
 CLOUD_STORAGE_ENDPONT_URL=http://localhost:4443   # used for GCS cloud storage emulator, leave empty for production
-# CLOUD_STORAGE_BUCKET_ID=resolution-client-assets
\ No newline at end of file
+# CLOUD_STORAGE_BUCKET_ID=resolution-client-assets
+
+# address of Puppeter local docker container, leave empty for production
+PUPPETEER_WS=wss://chrome.browserless.io?token=094632bb-e326-4c63-b953-82b55700b14c
```
DancingAxolotl, line 21: Are these real tokens? Is it okay to publish them?

README.md
```diff
@@ -361,3 +361,28 @@ docker run -d --name fake-gcs-server -p 4443:4443 -v ${PWD}/storage:/data fsouza
...
+Add the following line to *dev.env* configuration and restart the application:
+```
+PUPPETEER_WS=ws://localhost:3000
+```
+
+To connect to Browserless SAAS from localhost use the following URL:
+```
+PUPPETEER_WS=wss://chrome.browserless.io?token=76544be7-a2ae-4c6a-8d45-71a8ee183ded
+```
+where *token* is your API key
```
DancingAxolotl, line 388: Do we need to replace the token with a placeholder?

DancingAxolotl (changes requested):
Looks good overall. 
Just a few questions for clarification

iurevych:
Closing it since this project was deprioritised.

PR #176, opened 2022-10-28
feat: filter dead wallet addresses

enaqx:


src/types/common.ts
```diff
@@ -97,3 +97,8 @@ export type AllDomainTlds =
 export const UnsupportedTlds: Array<
   EvmUnstoppableDomainTlds | ZilliqaUnstoppableDomainTlds
 > = [UnstoppableDomainTlds.Coin, UnstoppableDomainTlds.Unstoppable];
+
+export type WalletAddress = string;
+export const DeadAdresses: Array<WalletAddress> = [
+  '0xdeaddeaddeaddeaddeaddeaddeaddeaddeaddead',
```
enaqx, line 103: @sammyluo Please let me know if we need to add any other addresses here.

src/controllers/DomainsControllerGetDomainsList.test.ts
```diff
@@ -601,6 +601,102 @@ describe('DomainsController', () => {
       });
       expect(res.status).eq(200);
     });
+
+    it('should not return domains with dead wallet addresses', async () => {
```
sammyluo, line 605: (Optional) It would be great if there's a test to cover the case that there are multiple dead wallet addresses.

sammyluo (approved):
LGTM! 🚀 ⭐

PR #183, opened 2022-11-02
Fix: case sensitive get records for owner

sammyluo:
Fix: case sensitive get records for owner


crazeekennee (approved):
LGTM !  I like creating the index on the new search use-case.

PR #200, opened 2022-11-17
[DT-926] Feature: Adding Heap tracking events in domain controller

DChan0319:
## Background

Want to start creating resolution service tracking events in order to determine who is using what endpoints.

## Changes

Included Heap tracking events in the domain controller.

## Code Review

## Confirmation

- [x] **Assignee Confirmation** - I (the author) have filled out the template above


DChan0319:
image

Tested the events coming through on staging. Just need to fix testing. Adding Heap broke the tests 😢

src/types/heap.ts
```diff
@@ -0,0 +1,12 @@
+export enum EHeapEvents {
+  DOMAIN_VIEWED = 'ecbe - viewed - domain',
```
sammyluo, line 2: I don't know if want to do `viewed`?

src/utils/heap.ts
```diff
@@ -0,0 +1,30 @@
...
+export const track = async (params: {
+  identity: string;
+  eventName: EHeapEvents;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  properties?: IHeapEventsProperties;
+  // This is used to de-dupe events and should not be shorter than 8 characters.
+  // Subsequent calls with same eventKey will not update data.
+  eventKey?: string;
+}) => {
+  const response = await request({
```
sammyluo, line 14: Let's check if `app_id` is provided first. If not, it should do nothing. The reason is Heap should be pluggable, meaning if someone wants host their own server without Heap, they should be able to do it.

sammyluo (changes requested):
Good start!

src/types/heap.ts
markthepixel: @unstoppable-allensun can weigh in here

DChan0319: Changed it to this format: `rsbe - GET - domain`

DChan0319:
image
Updated to include URI & agreed upon naming convention.

src/controllers/DomainsController.ts
```diff
@@ -58,7 +63,15 @@ export class DomainsController {
     };
 
     domainName = domainName.toLowerCase();
-
+    await track({
```
sammyluo, line 66: I think we should avoid `await` here. It blocks the process and if the event fails to be tracked the whole request will fail. It's better to use non-blocking approach here with `.catch` to log the error with `logger`.

It's also better to move the tracking to the end of the controller before the `return`, so that if anything goes wrong with the business logic, we don't fire false positive tracking events.

Same comment for the other trackings.

src/controllers/DomainsControllerGetRecords.test.ts
```diff
@@ -4,12 +4,14 @@ import { expect } from 'chai';
...
 
 describe('DomainsController', () => {
   let testApiKey: ApiKey;
 
   beforeEach(async () => {
     testApiKey = await ApiKey.createApiKey('testing key');
+    nockConfigure();
   });
 
   describe('GET /records', () => {
```
sammyluo, line 17: Let's add some tests to make sure heap tracking function is properly called when the controller is handling the request.

sammyluo (changes requested):

README.md
```diff
@@ -123,6 +123,7 @@ Note that the service is stateless, so the container doesn't need persistent sto
 | DEFAULT_MAX_REQUESTS                        | 5                                                 | :x:                | The default max number of requests before reaching the rate limit                                       |
 | METADATA_MAX_REQUESTS                       | 10                                                | :x:                | The max number of requests for the /metadata endpoints before reaching the rate limit                   |
 | WINDOW_MS                                   | 1000                                              | :x:                | The time window for requests before the window is reset for the rate limiting middleware                |
+| HEAP_APP_ID | - | :x: | The application ID used for sending trackig events |
```
sammyluo, line 126: Let's mention it's Heap application ID, and maybe add a hyper link to https://www.heap.io/

local.dev.env.sample
```diff
@@ -16,6 +16,7 @@ MORALIS_APP_ID=1
 METADATA_API=apikey
 OPENSEA_API_KEY=apikey
 RESOLUTION_APP_AUTH_KEY=test
+HEAP_APP_ID=developmentID
```
sammyluo, line 19: We can leave it as empty to avoid the tracking function being called. Otherwise it will throw a lot of errors during development

local.test.env.sample
```diff
@@ -16,3 +16,4 @@ MORALIS_API_URL=http://localhost
 MORALIS_APP_ID=1
 OPENSEA_API_KEY=test
 RESOLUTION_APP_AUTH_KEY=test
+HEAP_APP_ID=developmentID
```
sammyluo, line 19: We can leave it as empty to avoid the tracking function being called. Otherwise it will throw a lot of errors during development

src/controllers/DomainsController.ts
```diff
@@ -58,10 +63,19 @@ export class DomainsController {
     };
 
     domainName = domainName.toLowerCase();
-
     const supportedTLD = isSupportedTLD(domainName);
 
     if (!supportedTLD) {
+      track({
```
sammyluo, line 69: We can probably make it a function and call it whenever we need to avoid duplication

```suggestion
      const trackEvent = () => track({ ... });
      
      trackEvent();
```

sammyluo (approved):
LGTM! Great work! 🚀 ⭐

src/controllers/DomainsController.ts
```diff
@@ -58,17 +63,28 @@ export class DomainsController {
...
+        eventName: EHeapEvents.GET_DOMAIN,
+        properties: {
+          domainName,
+          apiKey: request.apiKey?.apiKey,
+          uri: request.originalUrl,
+        },
+      });
 
     if (!supportedTLD) {
+      trackEvent();
```
unstoppable-allensun, line 79: If these are being called in all code paths in this function, why not just do it once at the top of the function? (instead of calling it three times)

src/utils/heap.ts
```diff
@@ -0,0 +1,56 @@
...
+      }
+    } catch (error: any) {
+      logger.error(error);
+    }
+  }
+};
+
+// @see https://help.heap.io/definitions/properties/properties-overview/#property-data-types
+// https://help.heap.io/definitions/properties/how-to-create-defined-properties/
+export const normalizeHeapPropParam = (
```
unstoppable-allensun, line 40: Where is this being used?

src/types/heap.ts
```diff
@@ -0,0 +1,13 @@
+export enum EHeapEvents {
+  GET_DOMAIN = 'rsbe - GET - domain',
```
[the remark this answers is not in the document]
DChan0319: yes, not a problem!

src/utils/heap.ts
DChan0319: No where right now, i just wanted to have this ready to be used in the future, we are going to have POST requests. https://linear.app/unstoppable-domains/issue/DT-962/bulk-reverse-endpoint-on-resolution-service <-- is an example. We could probably use this function for the list of addresses that are being provided.

unstoppable-allensun: Sounds good! Thanks for the info.

src/controllers/DomainsController.ts
DChan0319: I think if the endpoint were to fail somewhere in between any of the logic it would still send a heap event somewhere regardless of the if statements, but the refactor is much cleaner anyways hahaha

src/utils/heap.ts
DChan0319: This can also be used for any GET request query params that are lists that we want to have an easier access to. instead of parsing the uri if it gets too long. There have been some partners that have been sending really long GET requests to search a butt load of things that would hit the character limit.... so even the uri custom param may need to be "broken up"

sammyluo (approved):
Beautiful! 🚀 ⭐

src/middleware/SendHeapEvent.ts
```diff
@@ -0,0 +1,37 @@
...
+import { Response, Request, NextFunction } from 'express';
+import { EHttpMethods } from '../types/common';
+import { track } from '../utils/heap';
+import { logger } from '../logger';
+
+@Middleware({ type: 'after' })
+export class SendHeapEvent implements ExpressMiddlewareInterface {
+  async use(req: Request, res: Response, next: NextFunction) {
+    switch (req.method) {
+      case EHttpMethods.GET:
```
MaximusMcCann, line 11: For what it's worth, as an engineer on another team, I don't know what an `EHttpMethods`.  E for event?  Just food for thought as UD grows and we have more new engineers looking at code :)

src/types/common.ts
```diff
@@ -117,3 +117,15 @@ export type WalletAddress = string;
 export const DeadAdresses: Array<WalletAddress> = [
   '0xdeaddeaddeaddeaddeaddeaddeaddeaddeaddead',
 ];
+
+export enum EHttpMethods {
```
MaximusMcCann, line 121:
```suggestion
export enum EventHttpMethods {
```

Maybe just spell it out, unless `E` doesn't actually stand for `Event` 😬 .
Or `TrackingEventHttpMethods` or `HeapHttpMethods` 🤷🏻‍♂

src/utils/heap.ts
```diff
@@ -0,0 +1,56 @@
...
+  let result = `${property[0]}`;
+
+  // Add data to string only if data does not pass character limit constraint.
+  if (property.length > 1) {
+    property.slice(1).forEach((prop) => {
+      if (`${result},${prop}`.length <= HEAP_PROP_MAX_CHARACTER_LIMIT) {
+        result = `${result},${prop}`;
+      }
+    });
+  }
```
MaximusMcCann, line 53: @unstoppable-allensun Here they've decided to straight truncate.  Thoughts?  I believe we decided to not convert the json to a truncated string, no?

unstoppable-allensun: Yah he said this was not being used anyways currently when I asked about this method.  This isn't truncating json though. It's converting arrays into comma separated list and truncating that. For that we could add "..." to the end to indicate it was truncated and then log that data was truncated.

src/middleware/SendHeapEvent.ts
DChan0319: E stands for Enum and I for Interface. When importing a type, i rather have some sort of indication as to what custom type, interface, or enum was created, right now it's not clear what i'm importing from the different files, i'd have to check and make sure in the type declaration with this i know I'm going to import and enum vs an interface or type. Totally fine and happy to remove though. Just thought it would be even clearer as a standard to our type definitions if introduced.

src/types/common.ts
DChan0319: Doesnt stand for Event. Stands for Enum HttpMethods, when importing it has a clear indication that these are enums vs an Interface. I could have an Interface with the same name and same description. EHttpMethods as this enum with the strings and an interface as IHttpMethods as another declared custom type.

src/utils/heap.ts
MaximusMcCann: oh 🤦🏻

src/types/common.ts
MaximusMcCann: Gotcha.  More common practice would be to just have a singular Enum them,  so `HttpMethod`

src/utils/heap.ts
```diff
@@ -0,0 +1,56 @@
...
+): string => {
+  const HEAP_PROP_MAX_CHARACTER_LIMIT = 1024;
+  let result = `${property[0]}`;
+
+  // Add data to string only if data does not pass character limit constraint.
+  if (property.length > 1) {
+    property.slice(1).forEach((prop) => {
+      if (`${result},${prop}`.length <= HEAP_PROP_MAX_CHARACTER_LIMIT) {
+        result = `${result},${prop}`;
+      }
```
MaximusMcCann, line 51: We should throw a log if it goes over and have the log include the entire `property`

PR #201, opened 2022-11-20
Display only supported tlds

sudoryan:
Fixes issue where metadata endpoint was displaying unsupported TLDs


src/controllers/MetaDataController.ts
```diff
@@ -356,6 +360,24 @@ export class MetaDataController {
     domainOrToken: string,
   ): Promise<OpenSeaMetadata> {
     const name = domainOrToken.includes('.') ? domainOrToken : null;
+
+    if (
+      name &&
+      !isSupportedTLD(name) &&
+      // We still want to return metadata for .coin domains
+      !belongsToTld(name, UnstoppableDomainTlds.Coin)
```
iurevych, line 368: To avoid issues in the future where we'd forget to add more endings here I'd also introduce a `isDeprecatedTld` helper.

iurevych (approved):
LGTM

PR #205, opened 2022-11-23
Add subdomains attribute type

sudoryan:
Adds subdomains attribute type

MetadataController.getAttributeType is now async

Screen Shot 2022-12-08 at 8 42 18 AM
Will display "subdomains" attribute next to length and level attributes to show the # of subdomains issued by a domain

### DIFF
--- a/src/models/Domain.ts
+++ b/src/models/Domain.ts
@@ -112,6 +112,15 @@ export default class Domain extends Model {
     return this.reverseResolutions.length > 0;
   }
 
+  static async getChildrenCountByParentName(
+    name: string,
+    repository: Repository<Domain> = this.getRepository(),
+  ): Promise<number> {
+    return repository.count({
+      where: { parent: { name } },
+      relations: ['parent'],
+    });
+  }
   static async findAllByNodes(
     nodes: string[],
     repository: Repository<Domain> = this.getRepository(),
--- a/src/models/Domain.ts
+++ b/src/models/Domain.ts
@@ -118,6 +118,7 @@ export default class Domain extends Model {
   ): Promise<boolean> {
     const child = await repository.findOne({
       where: { parent: { name } },
+      relations: ['parent'],
     });
     return Boolean(child);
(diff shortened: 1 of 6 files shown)


src/models/Domain.ts
```diff
@@ -112,6 +112,15 @@ export default class Domain extends Model {
     return this.reverseResolutions.length > 0;
   }
 
+  static async getChildrenCountByParentName(
+    name: string,
+    repository: Repository<Domain> = this.getRepository(),
+  ): Promise<number> {
+    return repository.count({
```
sammyluo, line 119: If the purpose is to figure out if a domain has a child so that we can determine if it has subdomain or not, then it's better to use `findOne` instead of `count` for better performance. We should also add caching on it using the `cache` option in TypeORM.

src/utils/metadata/index.ts
```diff
@@ -35,6 +35,7 @@ export enum AttributeType {
   Subdomain = 'subdomain',
   Animal = 'animal',
   Premium = 'premium',
+  Clean = 'clean', // domain has no subdomains
```
sammyluo, line 38: Just curious if "clean" is a good naming to represent a domain has no subdomain? 🤔

sammyluo (changes requested):
Good start! Left a change request.

src/utils/metadata/index.ts
sudoryan: It was suggested by Matt G. and intended to identify domains that are eligible for returns. The original suggestion was "clean (no subdomains)" 😅

iurevych:
@sudoryan should we add the new attribute type without modifying the default standard type? This is to avoid OpenSea caching issue where after this change all tokens will keep showing stale metadata (standard not clean type).

I'd propose adding a new type `WithSubdomains` if the domain has subdomains. This way as subdomains gain more popularity we'll start seeing this new trait becoming visible at OpenSea.

PR #208, opened 2022-11-28
Fix: Added resolution domain id indexes

DancingAxolotl:
## Background
Domain resolution cause a full table scan because domain ids are not indexed which degrades DB performance.

## Changes
 - Added indexes to `DomainsResolution#domainId` and `DomainsReverseResolution#domainId`


crazeekennee (approved):
Always in favor of better indexing !  LGTM.

sammyluo (approved):
Good catch! 🚀 ⭐

tunguyennnnn (approved):

PR #210, opened 2022-11-29
[DT-936] Added typeorm in memory cache to all GET requests DB queries

DChan0319:
## Background

We want to implement some basic forms of caching first in order to improve response times for the resolution service. 

## Changes

Added typeorm in memory caching implementation for all GET requests utilizing DB queries.

## Code Review

- [x] **Implementation** - satisfied acceptance criteria
- [x] **Communication** - notified engineers/stakeholders about impactful changes
- [x] **Documentation** - wrote readable code & commits, documented unintuitive code
- [x] **Testing** - unit & E2E test coverage, tested in staging, DB migrations backwards compatible
- [x] **Performance** - optimized expensive algorithms & database queries

## Confirmation

- [x] **Assignee Confirmation** - I (the author) have filled out the template above
- [x] **Reviewer Confirmation** - I (a/the reviewer) confirm 1) the author has filled out the template and checked the box that says **Assignee Confirmation** 2) I reviewed the code and selected all applicable items in the code review section.


src/controllers/DomainsController.ts
```diff
@@ -191,7 +193,9 @@ export class DomainsController {
     }
 
     qb.take(query.perPage + 1);
-    const domains = await qb.getMany();
+    const domains = await qb
+      .cache(env.CACHE.IN_MEMORY_CACHE_EXPIRATION_TIME)
```
sammyluo, line 197: What happens if `IN_MEMORY_CACHE_DISABLED` is set to `false`? Is it still going to be cached but not using in-memory caching mechanism?

src/controllers/DomainsController.ts
```diff
@@ -278,7 +282,9 @@ export class DomainsController {
       block_number: 'DESC',
       log_index: 'DESC',
     });
-    const lastTransferEvents = await domainEvents.getMany();
+    const lastTransferEvents = await domainEvents
```
sammyluo, line 285: Sorry, I should've clarified in the story, but I think this one would be an exception since it's supposed to return the latest blockchain events, so we should return real-time data.

src/database/connect.ts
```diff
@@ -18,6 +19,13 @@ if (!manager.connections.length) {
     maxQueryExecutionTime: -1,
     migrationsTableName: 'typeorm_migration',
     migrationsRun: runningMode.includes('MIGRATIONS'),
+    ...(!env.CACHE.IN_MEMORY_CACHE_DISABLED && {
```
sammyluo, line 22: Any reason we need to define the cache provider here again if it's ready in ormconfig.ts?

src/models/Domain.ts
```diff
@@ -135,19 +138,21 @@ export default class Domain extends Model {
...
-          cache: cache ? 600000 /* 10 mins */ : undefined,
+          cache: cache ? env.CACHE.IN_MEMORY_CACHE_EXPIRATION_TIME : undefined,
         })
       : undefined;
   }
 
   static async findOrBuildByNode(
     node: string,
     repository: Repository<Domain> = this.getRepository(),
+    cache?: boolean,
```
sammyluo, line 149: I don't think we should provide a cache option for find or build operation. Imagine if some code that looks up the domain first and it doesn't exist, then it returns a new Domain instance and persisted into the DB. Then if a function looks up the same data within 10 mins, it will hit the cache and still say that the domain doesn't exist.

sammyluo (changes requested):
Great start! Left some minor change requests.

src/database/connect.ts
DChan0319: When i was running the tests locally, I wasn't seeing the caching actually taking place, but the caching was throwing errors when I added it here. My question i guess is when or how does this config get used or read?

src/controllers/DomainsController.ts
DChan0319: If caching is not disabled, many of the unit tests were failing. As soon as i disabled the cache, the unit tests pass. So, to me that meant, if caching is disabled, everywhere were we see `.cache()` format of code doesn't actually take place unless in the config it is provided. The only place where test results were being cached and being affected was when i enabled caching in the connect.ts file.

src/database/connect.ts
DChan0319: The only place i see ormconfig.ts being referenced is in the tsconfig.json file.. so i dont think it's even being used other than being scanned for typescript.

sammyluo: It should be picked up by TypeORM automatically

DChan0319: ah, okay i see `ts-node` in the scripts allows for this. I'll remove it here then. I don't think we even need the env.CACHE.IN_MEMORY_CACHE_DISABLED environment variable then either cause it was only introduced for testing and testing doesn't need this if i remove this logic.

src/controllers/DomainsController.ts
DChan0319: In the revision, the cache should never be disabled. I only introduced the flag because i thought i needed to turn it off for tests.

src/models/Domain.ts
```diff
@@ -273,10 +276,12 @@ export default class Domain extends Model {
     name: string,
     blockchain: Blockchain,
     repository: Repository<Domain> = this.getRepository(),
+    cache?: boolean,
   ): Promise<Domain> {
     const domain = await repository.findOne({
       where: { name },
       relations: ['resolutions', 'reverseResolutions', 'parent'],
+      cache: cache ? env.CACHE.IN_MEMORY_CACHE_EXPIRATION_TIME : undefined,
```
sammyluo, line 284: Same comment as find or create operation - we shouldn't provide cache option.

src/database/connect.ts
```diff
@@ -3,6 +3,7 @@ import { WinstonTypeormLogger } from './WinstonTypeormLogger';
 import SnakeNamingStrategy from './SnakeNamingStrategy';
 import { logger } from '../logger';
 import { env } from '../env';
+import InMemoryCacheProvider from './TypeormInMemoryCache';
```
sammyluo, line 6: Unused variable.

sammyluo (approved):
Great work! 🚀 ⭐

PR #221, opened 2022-12-07
Fix: clean up hanging promises and reorg handling

DancingAxolotl:
## Background
Fixes reorg handling issues

## Changes
 - Added `no-floating-promises` to eslint
 - Fixed floating promises
 - Checked for undefined resolutions when rebuilding a domain during reorg handling


src/middleware/SendHeapEvent.ts
```diff
@@ -14,15 +14,15 @@ export class SendHeapEvent implements ExpressMiddlewareInterface {
             throw new Error('No Heap Event Name was set in controller logic');
           }
 
-          track({
+          void track({
```
DancingAxolotl, line 17: For now I assumed that we are explicitly not waiting for this promise so that event tracking doesn't increase response latency.

@DChan0319 Since you added heap tracking, could you check this to make sure we don't need to await this promise?

src/workers/eth/EthUpdater.ts
```diff
@@ -773,9 +782,12 @@ export class EthUpdater {
...
 }
 
-export function startWorker(blockchain: Blockchain, config: any): void {
+export function startWorker(
+  blockchain: Blockchain,
+  config: EthUpdaterConfig,
+): void {
   if (config.RESYNC_FROM !== undefined) {
-    new EthUpdater(blockchain, config).resync().then(() => {
+    void new EthUpdater(blockchain, config).resync().then(() => {
```
SteversIO, line 790: Interesting use of void inline.   My understanding is that it's redundant.  It returns undefined after executing the full inline expression.

Does the use of void here actually change the business logic?

tunguyennnnn: no, it's typescript thing. it just says that we are not awaiting for the promise because of the new eslint rule

PR #245, opened 2022-12-19
Fixed validation error message, Added validation tests to domains con…

tunguyennnnn:
…trollers

## Background

Input validation was recently introduced. But the input message needs to be improved for invalid characters.

## Changes

* Changed error message for invalid input such as `+name+.crypto`
* Added more tests to domains endpoints

## Code Review

## Confirmation


src/middleware/inputValidators/validateDomainNameQuery.ts
```diff
@@ -55,11 +59,11 @@ export function ValidateAndTransformOnDomainName(
       throw new InvalidInputError('Empty input');
     }
 
-    if (
-      !value.includes('.') ||
-      !isValidDomainNameLabel(value) ||
-      !isSupportedTLD(value)
-    ) {
+    if (!value.includes('.') || !isValidDomainNameLabel(value)) {
```
DancingAxolotl, line 62: Not sure if it was like this from the start, but a TLD will not pass this check
So
https://metadata.unstoppabledomains.com/metadata/crypto
Will return an error because it does not contain a `.`

TLDs are technically valid domains, so I think we should return results for them too.

src/controllers/CommonTestSuites.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  includeTokenTests: boolean;
+}) {
+  const {
+    getPath,
+    isAuthRequired = false,
+    includeDomainNameTests,
+    includeTokenTests,
+  } = options;
+
+  describe(`Validate input: ${getPath(':domainOrToken')}`, () => {
```
DancingAxolotl, line 20: It looks like these are tests for the middleware, maybe we can add them in `src/middleware/inputValidators.test.ts`?

src/middleware/inputValidators/validateDomainNameQuery.ts
tunguyennnnn: ✅ Added back the behavior, this maybe a regression at somepoint.

src/controllers/CommonTestSuites.test.ts
tunguyennnnn: added

src/controllers/DomainsControllerGetDomainTransfers.test.ts
```diff
@@ -57,6 +59,13 @@ describe('DomainsController', () => {
       }
     }
 
+    commonParamsValidatorTestSuite({
+      getPath: (domain: string) => `/domains/${domain}/transfers/latest`,
+      isAuthRequired: true,
+      includeDomainNameTests: true,
+      includeTokenTests: false,
+    });
```
DancingAxolotl, line 67: Since the validator is already covered by unit tests, do we need to re-check them here? Can we just check that the validator is called with a mock? 
This might save some time on tests since we don't have to run the same suite multiple times.

DancingAxolotl: And same for other instances of `commonParamsValidatorTestSuite`

DancingAxolotl (approved):
LGTM! Nice tests 💪
Left a minor question, but it doesn't block the PR.

PR #259, opened 2022-12-27
[DT-984] fix: Updated migration query to fix subdomain parent ids

DChan0319:
## Background

Need to fix the subdomain parent ids that are incorrect on prd

## Changes

Added migration script to be ran during build

## To Do

- [ ] Verify subdomain parent ids are correct once migration is ran
## Code Review

This Pull Request was thoroughly reviewed and meets all Code Review Standards pertaining to:

- [ ] **Implementation** - satisfied acceptance criteria
- [ ] **Communication** - notified engineers/stakeholders about impactful changes
- [ ] **Error handling** - handled, logged & alerted on errors
- [ ] **Documentation** - wrote readable code & commits, documented unintuitive code
- [ ] **Testing** - unit & E2E test coverage, tested in staging, DB migrations backwards compatible
- [ ] **Security** - sanitized inputs, vetted dependencies, validated/secured API requests, no committed secrets
- [ ] **Performance** - optimized expensive algorithms & database queries

Please select all **applied** standards relevant to this PR.
## Confirmation

- [ ] **Assignee Confirmation** - I (the author) have filled out the template above
- [ ] **Reviewer Confirmation** - I (a/the reviewer) confirm 1) the author has filled out the template and checked the box that says **Assignee Confirmation** 2) I reviewed the code and selected all applicable items in the code review section.

PR #261, opened 2022-12-28
Revert "[DT-984] fix: Updated migration query to fix subdomain parent ids"

DChan0319:
Reverts unstoppabledomains/resolution-service#259


DChan0319:
Reverting as it was unsuccessful in migrating to PROD and it's blocking future PRs to be deployed successfully. Safe to revert.

PR #268, opened 2022-12-31
[DT-964] feat: Resolution service common worker framework 

DancingAxolotl:
## Background
 - See linear task

## Changes
 - Major refactor of workers to extract common functionality and make them more extendable 
 - Integration/e2e tests of workers are now optional with an env variable
 - Added worker runner and factory to streamline setting up workers

## To Do

A couple major TODOs:
- [x] add unit tests for new code
    - [x] For now legacy tests are kept in place with minimal changes and can be used to verify workers. But we should reorganize these tests and cover new parts with separate unit tests. 
- [x] Get rid of old Updater/Worker classes. With the new architecture we can introduce a common factory to create worker instances. For now old classes are kept for unit tests. 
- [x] Test on staging before releasing to prod.

## Code Review

## Confirmation


src/workers/workerFramework/IWorker.ts
```diff
@@ -0,0 +1,4 @@
+export interface IWorker {
```
DChan0319, line 1: I like the whole 'I' + 'variableName' naming convention to add to the front of Interface naming and 'E' + 'variableName' for enums. I tried introducing that somewhere my first month or so, but got flagged because it was not an org wide naming convention that's followed. I think it's distinct and clear tho as to what you're trying to import as a interface, enum, etc.

src/workers/workerFramework/WorkerRepository.ts
```diff
@@ -0,0 +1,457 @@
...
+
+  private get context(): WorkerRepositoryTxContext {
+    const context = WorkerRepository.txContexts[this.blockchain];
+    if (!context) {
+      throw Error('transaction closed');
+    }
+    return context;
+  }
+
+  convertToArray<T>(val: T | T[]): T[] {
```
DChan0319, line 56: Can't we have this as a utility function to be used repo wide if needed, instead of just the class? Don't see a reason to contain it in here.

src/workers/workerFramework/WorkerRepository.ts
```diff
@@ -0,0 +1,457 @@
...
+  private constructor(blockchain: Blockchain, networkId: number) {
+    this.blockchain = blockchain;
+    this.networkId = networkId;
+    this.logger = WorkerLogger(blockchain);
+  }
+
+  private get context(): WorkerRepositoryTxContext {
+    const context = WorkerRepository.txContexts[this.blockchain];
+    if (!context) {
+      throw Error('transaction closed');
```
DChan0319, line 51: Can we have a more defined Error structure for our errors here?

src/types/common.ts
```diff
@@ -129,3 +129,33 @@ export enum HttpMethods {
...
+  'ApprovalForAll',
+] as const;
+
+export const ContractManagementTypes = [
+  'NewURIPrefix',
+  'Upgraded',
+  'AdminChanged',
+] as const;
+
+export const EventTypes = [
```
DChan0319, line 155: Can we have these as enums? And then in workerEvent have the type as one of the Enum values instead of a string?

tunguyennnnn (approved):

tunguyennnnn:
I added a PR to address some comment I had https://github.com/unstoppabledomains/resolution-service/pull/270/files

src/workers/workerFramework/IWorker.ts
DancingAxolotl: Interesting, I don't think I've seen many interfaces used in our code in general, especially as public interfaces that are meant to be implemented by third party devs.

I think I got the name from some old habits from Java/C++. So it might not be the best option. Maybe it is worth getting some feedback org-wide for the naming convention here.

src/workers/workerFramework/WorkerRepository.ts
DancingAxolotl: That's a great point

In general the worker framework should be able to handle generic errors in case the strategy throws them, but we should totally have a framework-specific error type.

src/workers/eth/UnsWorkerStrategy.ts
```diff
@@ -0,0 +1,441 @@
...
+        `Fetched ${
+          cnsEvents.length
+        } cnsEvents from ${fromBlock} to ${toBlock} by ${
+          toBlock - fromBlock + 1
+        } `,
+      );
+    }
+
+    // Merge UNS and CNS events and sort them by block number and index.
+    const events: Event[] = [...cnsEvents, ...unsEvents];
```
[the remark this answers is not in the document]
DancingAxolotl: They aren't dependent on each other. But also they are pretty much the same. So the handlers would be shared between them. There would be either lots of code duplication, or a shared class/module with event handling code.

So I think it is easier to just keep them in one worker. Especially since this is technically the same nameservice and domains may migrate from CNS to UNS.

src/workers/eth/UnsWorkerStrategy.ts
```diff
@@ -0,0 +1,441 @@
...
+          }
+          case 'SetReverse': {
+            await this.processSetReverse(event);
+            break;
+          }
+          case 'RemoveReverse': {
+            await this.processRemoveReverse(event);
+            break;
+          }
+          case 'Approval':
```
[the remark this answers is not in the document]
DancingAxolotl: Not really. They just got copy pasted from the original implementation. 

It might be worth flagging explicitly which events we are ignoring, but they will get saved to the DB anyway. So I'd say it is up to the strategy implementation.

sammyluo (commented):
Great start! 🚀 ⭐ Left some minor comments and questions.

src/workers/workerFramework/IWorker.ts
SteversIO: I think we should agree on a styleguide as a team and try hard to adhere to it. The prefix `I` is not widely used in the style guides I'm accustomed to using.  Many of the teams I joined did not have a standard, so we defaulted to Google's style guide

- [JavaScript](https://google.github.io/styleguide/jsguide.html)
- [Java](https://google.github.io/styleguide/javaguide.html)

.NET uses them; Microsoft tends to stray far from the rest of the open source community. (and I'm not a fan of Microsoft's products or code quality pre-Satya Nadella; they get away with a lot of bad practices because their sales are mostly enterprise)

src/workers/workerFramework/BaseWorker.ts
```diff
@@ -0,0 +1,251 @@
...
+    const [firstNetBlock, lastNetBlock] = await Promise.all([
+      this.workerStrategy.getBlock(latestEventBlocks[0].blockNumber),
+      this.workerStrategy.getBlock(
+        latestEventBlocks[latestEventBlocks.length - 1].blockNumber,
+      ),
+    ]);
+
+    // If the oldest event block doesn't match, the reorg must be too long.
+    if (firstNetBlock.blockHash !== latestEventBlocks[0].blockHash) {
+      throw new EthUpdaterError(
```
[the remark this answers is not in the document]
SteversIO: 👍  +1

src/database/migrations/1672504312180-ZNSStatusSwapAtxuidAndBlockNum.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+import { WorkerStatus } from '../../models';
+import { Blockchain } from '../../types/common';
+
+export class ZNSStatusSwapAtxuidAndBlockNum1672504312180
+  implements MigrationInterface
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `UPDATE "resolution_worker_status" SET last_mirrored_block_number=last_atxuid WHERE location='ZIL'`,
```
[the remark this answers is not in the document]
DancingAxolotl: This migration saves the `last_atxuid` as a `last_mirrored_block_number` for `ZIL`. 

Historically ZIL used a different column in the worker status table to save its status - `last_atxuid`. Technically `last_atxuid` is not a block number, so it was separate. But now that we have a common framework we want to save worker statuses consistently to the `last_mirrored_block_number` column. We copy the value here just so we don't start syncing from scratch once this is merged.

We can consider renaming the `last_mirrored_block_number` column too. Since it is now more like an identifier that consistently increases (e.g. block height, atxuid, or some other number for exotic structures like DAGs, shards, etc.).

src/workers/workerFramework/Types.ts
```diff
@@ -0,0 +1,109 @@
+import { Blockchain } from '../../types/common';
```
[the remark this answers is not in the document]
DancingAxolotl: If you are talking about the event type enums, I don't think it should be a part of the framework. Event types are part of the nameservice implementation. We only store the type as a generic string in our DB since different nameservices may have different events. Each worker implementation should define its own event types which it can parse.

Or did you mean another enum?

src/workers/workerFramework/Types.ts
```diff
@@ -0,0 +1,109 @@
...
+  // Blockchain id of resolution
+  blockchain: Blockchain;
+  // Network id of resolution
+  networkId: number;
+  ownerAddress?: string | null = null;
+  resolver?: string | null = null;
+  registry?: string | null = null;
+  resolution?: Record<string, string> = undefined;
+
+  get isUnchanged() {
```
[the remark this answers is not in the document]
DancingAxolotl: I agree, this one is a bit tricky to understand. 

Basically, resolution changes are implemented in an "upsert" manner. So an "empty" resolution means that no changes were made and there is nothing to "upsert". This is used for some internal optimisation in the `WorkerRepository` to skip writing resolutions that weren't changed which leads to the naming `isUnchanged`.

We could change it to something like `dirty` or `updated`. This might be easier to understand.

src/workers/eth/EthUpdater.ts
```diff
@@ -1,791 +1,48 @@
...
 import { Blockchain } from '../../types/common';
 import { EthUpdaterConfig } from '../../env';
-import winston from 'winston';
-import { cacheSocialPictureInCDN } from '../../utils/socialPicture';
+import { BaseWorker } from '../workerFramework/BaseWorker';
+import { UNSWorkerStrategy } from './UnsWorkerStrategy';
+import { WorkerConfig } from '../workerFramework';
 
+// Keep for legacy tests compatibility
 export class EthUpdater {
```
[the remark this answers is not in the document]
DancingAxolotl: Ultimately the we will either use the `BaseWorker` directly since it is not an abstract class or extend it to override some specific functionality.

I didn't want to extend the base worker here as these classes are kept just to have the old unit tests working. We should get rid of them once we have proper unit tests and an entity to manage worker. Extending the base worker would still break the interface used by the unit tests, so it's not worth it.

src/types/common.ts
DancingAxolotl: After looking into it a bit more, there a couple reasons why I don't think we should make this into an enum. 

First, for the general event entity. These events are specific to UNS/CNS nameservices. We will not be able to reuse these enums for other nameservices and we shouldn't limit our entity to these specific event types. Ultimately we don't care about the actual event type or its parameters. We just save events for accountability and for restoring domains during reorg handling.

And second, for CNS/UNS specifically. In this case we use the arrays to determine if an event belongs to one of these groups (e.g. if it has a `node` value). Which is useful for event parsing. So for our UNS implementation we could make these enums, but we don't really use them as enums.

Ultimately the way event types are parsed is up to the strategy implementation, we shouldn't limit it on the framework side. And on the UNS implementation side they are more useful as arrays. So I think it's not worth it to turn these into enums but maybe there's something I'm missing here.

src/workers/workerFramework/Types.ts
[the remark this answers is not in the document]
DancingAxolotl: Also note that this is exclusively a `get` field. Clients can't write to this field. So even if it's not super clear, they can't mess up the entity by setting an incorrect value.

src/types/common.ts
SteversIO: +1 to this. Also makes me wonder if `common` is the right location for this type of variable declaration.

As an aside, in general when you find code with a long chain of switches or if/else if, you'll find that polymorphism does a much better job to reduce code complexity, improve readability and increase modularity.

src/workers/eth/CnsResolver.ts
```diff
@@ -47,24 +48,23 @@ export class CnsResolver {
...
+        BigNumber.from(resolution.node),
       );
     } catch (error: any) {
       if (!error.message.includes(InvalidValuesError)) {
         throw error;
       }
     }
 
-    return {};
+    return undefined;
```
SteversIO, line 67: Is there a specific reason why sometimes `undedined` is returned in code while other places a `null` is returned?

src/workers/eth/CnsResolver.ts
```diff
@@ -164,13 +164,24 @@ export class CnsResolver {
...
 
+    if (Object.keys(records).length === 0) {
+      return undefined;
+    }
     return records;
   }
 
+  static normalizeResolver(resolver: string | null | undefined): string | null {
+    if (!resolver) {
+      return null;
```
SteversIO, line 175: Here's an example of a `null` returned whereas other places it's `undefined`.

src/workers/eth/UnsWorkerStrategy.ts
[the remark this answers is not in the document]
SteversIO: Literally started a comment asking the same thing.  Good catch everyone 💪🙏🚀. Maybe we rename the worker to something more encapsulating of the two?  `UberWorker` ?

Also, since Ethereum and polygon have Uns and CNS to deal with, is there a potential head of the blockchains The similar issue that I have to support multiple to my contracts concurrently?

SteversIO (commented):
General questions

src/workers/workerFramework/IWorker.ts
```diff
@@ -0,0 +1,4 @@
...
+  run(): Promise<void>;
+  resync(fromBlock: number): Promise<void>;
+}
```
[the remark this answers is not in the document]
DancingAxolotl: They are optional since not many networks "suffer" from reorgs as much as Polygon does. And reorg checking takes a few RPC calls. So just to save some performance overhead you can disable them.

But in either case, reorg handling is done automatically in the `run` function. There's no need for an interface function for reorg handling.

src/workers/eth/CnsResolver.ts
DancingAxolotl: Resolvers were a separate smart contract for CNS that stored resolution records.
As such it may have a `NULL` value, meaning "no resolver is set for this domain".

src/types/common.ts
DancingAxolotl: Yeah, common becomes confusing here. It could be common for the worker framework, or it could be common for our UNS/CNS/ZNS workers or the API. I think it will require a more in-depth refactor to separate out different components of the resolution service.

As for using polymorphism for event types. That is a nice idea. I'd say it is up to the worker strategy implementation at the moment. But we may setup some utils in the framework to write individual event handlers in the future.

In our case the refactor is already pretty big, so I didn't want to change it much more than we planned initially. It will impact more of the code and will require more testing, debugging, etc.

src/workers/eth/UnsWorkerStrategy.ts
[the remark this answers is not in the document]
DancingAxolotl: The plan was to have one worker per nameservice (UNS, ZNS). So if multiple nameservices live on the same blockchain they would have different workers since they may have different events, data structures, etc.

At this point it's not really UNS+CNS. It is more like "UNS with some legacy CNS functions". Since CNS is basically a subset of UNS, I think splitting them is not worth it.

src/workers/WorkerRunner.ts
```diff
@@ -0,0 +1,42 @@
+import { setIntervalAsync } from 'set-interval-async/dynamic';
+import { logger } from '../logger';
+import { IWorker } from './framework';
+
+export type RunOptions = { resyncFrom?: number; fetchInterval?: number };
+
+export default class WorkerRunner {
+  workers: {
```
tunguyennnnn, line 8:
```suggestion
  workers: Array<{...
```
nit: make it more obvious.

src/workers/eth/CnsResolver.ts
```diff
@@ -25,11 +23,13 @@ export class CnsResolver {
     },
     {} as Record<string, string>,
   );
+  private workerRepository: WorkerRepository;
 
-  constructor(config: CryptoConfig) {
+  constructor(config: CryptoConfig, workerRepository: WorkerRepository) {
```
tunguyennnnn, line 28: 💯 for passing repository here

src/workers/eth/CnsResolver.ts
tunguyennnnn: 🤔 , well i had enough debate about this in the past. and code base i worked with before just always return undefined for no value or null value unless it's for client consumption.

tunguyennnnn (approved):
really minor comments. look great!
Next time for big PR like this we should break it down to smaller PRs

tunguyennnnn (approved):

PR #300, opened 2023-03-07
[DT 1099] Add timeout promise wrapper to moralis outbound requests

tunguyennnnn:
## Background

Some requests to moralis takes too long to respond

## Changes

* Add a timeout of 500ms to terminate requests to moralis from Rest API calls

## Code Review

## Confirmation


src/services/MetadataService.ts
```diff
@@ -70,6 +71,9 @@ export class MetadataService {
 
   async fetchTokenMetadata(
     resolution: DomainsResolution,
+    config: {
```
sammyluo, line 74: Why not make it an optional config so you don't have to modify the other unrelated function calls?
```suggestion
    config?: {
      withTimeout?: boolean
```

Also what do you think of using a number for timeout instead of boolean to give the config power to the function? Something like
```
  fetchTokenMetadata(resolution, { timeout: 500});
```

tunguyennnnn: > Why not make it an optional config so you don't have to modify the other unrelated function calls?

good point 👍 

>Also what do you think of using a number for timeout instead of boolean to give the config power to the function? Something like

`fetchTokenMetadata` does a few network calls beside moralis. setting this as boolean is to tell the code that if `withTimeout` is true, each network service need to consider its own timeout policy.
Right now we only apply to moralis but we will need to look at other network call with different timeout config as well

sammyluo (approved):
LGTM

PR #308, opened 2023-03-20
[API-406] resolution api latency tweaks

DancingAxolotl:
## Background

Some minor tweaks to improve resolution service API p99 latency and latency spikes.

## Changes

 - Made API docs (swagger) into a separate running mode to improve startup latency for metadata API
 - Changed default running mode for `*-meta` instances to `METADATA_API`
 - ~Changed `/status` endpoint to use time since last update for status instead of block height to remove external queries~  moved to a separate PR #328 
 - Tweaked `/metadata` endpoint:
  -- Removed `count` for subdomains - use relation instead
  -- Batch DB queries internally to save postgres connections under high N of concurrent requests 
  -- Removed on-chain lookup for missing domains

## To Do

 - [x] test to make sure everything works - tested on staging, seems to work okay. Noticed significant [performance improvement](https://unstoppabledomains.slack.com/archives/C04KA8YNGQJ/p1678918841977239) for request spikes (>500req/s).
 - [x] fix unit tests
 - [x] clean up db query batching
 - [] ~set up secrets for new status timeouts, come up with reasonable estimates for these values (currently all set to 5 min)~

## Code Review

## Confirmation


src/services/DomainService.ts
```diff
@@ -1,29 +1,66 @@
...
-  let domain =
-    (await Domain.findByNode(tokenName, undefined, true)) ||
-    (await Domain.findOnChainNoSafe(tokenName));
+  const [domain, znsDomain] = await Promise.all([
+    await findByNodeQueue(tokenName),
+    ...(IsZilDomain(domainName)
+      ? [await findByNodeQueue(znsNamehash(domainName))]
+      : []),
+  ]);
+  const finalDomain = domain || znsDomain; // || await Domain.findOnChainNoSafe(tokenName);
```
DancingAxolotl, line 58: @sammyluo I think we discussed it before, but I'd like to bring it up again, `findOnChainNoSafe` is probably unnecessary here since we greatly reduced confirmation times for all chains. It also allows for DDoS attacks by spamming known non-existing token IDs (0, 1, 2, 3, etc.). So I think it would be better to drop it entirely.

However, keeping it is still an option if we need to. It will increase our average latency, but the main bottleneck was postgres connections so we should be okay.

src/services/DomainService.ts
```diff
@@ -1,29 +1,66 @@
...
+};
+
+const findByNodeQueue = async (node: string) => {
+  if (!processing) {
+    processing = true;
+    setImmediate(processQueue);
+  }
+  return new Promise<Domain | undefined>((resolve, reject) => {
+    processingQueue.push({ node, resolve, reject });
+  });
```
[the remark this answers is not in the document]
DancingAxolotl: Yes, normally you would need some synchronization to avoid situations like that. Thankfully, node.js is single-threaded. So that shouldn't be an issue.

Letting Node decide would be great, but the issue is typeorm which has a limited number of connections to the DB (10 currently). When we have, for example, 80 incoming requests at the same time, the first 10 will get their DB connections, while the rest 70 will be stuck waiting in the typeorm queue. At the same time another 80 requests will come in. So the total queue just keeps growing, resulting in huge latencies.

This way we let all 80 requests "push" their DB queries without eating up postgres connections. Once all incoming requests are handled, we can make a single DB query and return results to be handled on the next iteration of the event loop.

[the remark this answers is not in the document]
tunguyennnnn: what's this trying to do is trying to mimic a batch load pattern which is very common in graphql. I feel like we need to introduce proper interface or use existing library that already use this for better abstraction otherwise it will be hard to debug 
Have you looked into libraries such as this https://www.npmjs.com/package/batchloader to see if it can achieve what you need? (the `load` method  is the one that will perform query to DB in 1 tick)

or if you want to build your own this is a good reference.https://github.com/graphql/dataloader in https://github.com/graphql/dataloader/blob/main/src/index.js, it's somewhat similar to what you try to achive except that the assumption of the lib is queries are batched within 1 tick of the even loop. But regardless, it offers APIs batch load functions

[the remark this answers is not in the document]
tunguyennnnn: and since it operates on the assumption of 10 connections in the pools, it should be configurable on how much time per batch, how many items per batch.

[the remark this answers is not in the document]
DancingAxolotl: Oh, nice find @tunguyennnnn !!!
I was looking for a library like this, but for some reason I couldn't find it :)
It seems to do the same thing as my little implementation except with a `setTimeout` instead of `setImmediate` but ultimately it should have the same effect. I'll try it here. Thanks!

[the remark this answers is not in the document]
DancingAxolotl: `batchloader` worked. Thanks everyone!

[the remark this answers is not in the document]
tunguyennnnn: awesome!

src/services/DomainService.ts
```diff
@@ -1,29 +1,41 @@
...
+const domainsBatchLoader = new BatchLoader(
+  (_ids: string[]) =>
+    Domain.findAllByNodes(_ids, undefined, true, [
+      'resolutions',
+      'parent',
+      'children',
+    ]),
+  null,
+  0,
+  200,
```
anhan32, line 20: can you add comments to what these inputs correlate too? does this configuration essentially mean "for each loop, batch 200 requests"?

anhan32 (commented):
looks ok to me but defer @sammyluo @tirish @tunguyennnnn for approval

package.json
```diff
@@ -55,6 +55,7 @@
     "@zilliqa-js/crypto": "^2.3.0-alpha.0",
     "@zilliqa-js/util": "^2.2.0",
     "@zilliqa-js/zilliqa": "^2.3.0-alpha.0",
+    "batchloader": "^0.0.22",
```
sammyluo, line 58: I'm not sure about this package. Last published was 3 years ago so it seems like it's no longer being mainstream. Do you know any other alternatives?

src/api.ts
```diff
@@ -66,63 +66,68 @@ if (env.APPLICATION.BUGSNAG_API_KEY) {
   api.use(bugsnagPlugin?.errorHandler);
 }
 
-const schemas = validationMetadatasToSchemas({
-  refPointerPrefix: '#/components/schemas/',
-});
+if (
+  env.APPLICATION.RUNNING_MODE.includes('API') ||
+  env.APPLICATION.RUNNING_MODE.includes('API_DOCS')
```
sammyluo, line 71: Please rebase as you don't need this anymore

src/controllers/StatusController.ts
```diff
@@ -55,42 +46,29 @@ class StatusResponse {
...
-      return {
-        latestMirroredBlock,
-        latestNetworkBlock: -1,
-        networkId: config.NETWORK_ID,
-        acceptableDelayInBlocks: config.ACCEPTABLE_DELAY_IN_BLOCKS,
-        isUpToDate: false,
-      };
-    }
+    const workerStatus = await WorkerStatus.findOne({ location: blockchain });
+    const lastUpdated = workerStatus?.updatedAt?.getTime() || 0;
```
sammyluo, line 60: Not a problem now, but using the table `updatedAt` field might not correctly reflect the last successfully synced time, especially if we are expanding the table to keep track of other fields in the future, this will be a bug.

src/services/DomainService.ts
```diff
@@ -1,29 +1,41 @@
...
+    tokenName,
+    ...(IsZilDomain(domainName) ? [znsNamehash(domainName)] : []),
+  ]);
+  const finalDomain = domain || znsDomain; // || await Domain.findOnChainNoSafe(tokenName);
 
-  if (!domain && IsZilDomain(domainName)) {
-    domain = await Domain.findByNode(znsNamehash(domainName), undefined, true);
-  }
-  const supportedTLD = domain ? isSupportedTLD(domain.name) : false;
+  const supportedTLD = finalDomain ? isSupportedTLD(finalDomain.name) : false;
```
sammyluo, line 35: Now that you are here, we can even have another `isSupportedTLD` at the beginning of the function to do a simple check first. If it's not supported, then no need to query at all.

src/services/MetadataService.ts
```diff
@@ -299,7 +299,7 @@ export class MetadataService {
       },
       {
         trait_type: DomainAttributeTrait.Subdomains,
-        value: await Domain.getSubdomainCountByParentName(domain.name),
```
sammyluo, line 302: Might as well just delete the function `getSubdomainCountByParentName()` implementation?

sammyluo (changes requested):

package.json
DancingAxolotl: Originally I planned to implement this myself. @tunguyennnnn found this package and it seems to do the same thing.
It is a pretty simple utility, so it makes sense that it's not updated very often.
There is https://github.com/graphql/dataloader which seems to be updated more often, but I'm not sure if it will work the same way. I'll take a look.

src/controllers/MetaDataController.test.ts
```diff
@@ -516,7 +516,7 @@ describe('MetaDataController', () => {
       });
     });
 
-    it('should query the newURI event on the fly to return a freshly minted domain', async () => {
+    it.skip('should query the newURI event on the fly to return a freshly minted domain', async () => {
```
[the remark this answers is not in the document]
DancingAxolotl: I'll remove it completely if we decide to remove the "on the fly" fetching. See this comment:
https://github.com/unstoppabledomains/resolution-service/pull/308#discussion_r1142335677

package.json
tunguyennnnn: yea i don't remember https://github.com/graphql/dataloader was updated that often when I first saw it few years ago 😄.  but yea I'd not expect this package to be updated often. 

@sammyluo has a point that this package has only 36 stars on github, we can try to read through the code to see if it does what we need

src/services/DomainService.ts
DancingAxolotl: We need to find the domain before checking this in case we get a token id as input. If we only check in the beginning we will not catch unsupported token ids.

package.json
DancingAxolotl: Looks like `dataloader` is a bit more advanced. I was able to replace `batchloader` with it, just need to test to make sure that it works the same.

tunguyennnnn: yea i wouldn't use `dataloader`, it's specific to graphql to batch resolvers query within a single tick of the event loop. I just mentioned it as I was surprised that it's got maintained often because the original implementation was pretty niche already. `batchloader` is more for generic usage like ours

src/api.ts
DancingAxolotl: I left the env variable to allow us to disable swagger generation entirely to improve startup latency. For example, the metadata API does not need it, so we can save some time on startup.

src/controllers/StatusController.ts
tunguyennnnn: yea I wonder how much this would save us for not fetching from providers. We don't have much traffic for this endpoint

src/services/DomainService.ts
```diff
@@ -1,29 +1,52 @@
...
+      true,
+      ['resolutions', 'parent', 'children'],
+    );
+    return keys.map(
+      (key) => domains.find((d) => d.node === key) || new Error('not found'),
+    );
+  },
+  {
+    maxBatchSize: 200,
+    batchScheduleFn: setImmediate,
```
tunguyennnnn, line 25: 🤔 i am not sure about this. have you test actual batch size with using `setImmediate`, I think it would just execute immediately per request because each api requests coming will make load function executed immediately

src/services/DomainService.ts
```diff
@@ -1,29 +1,52 @@
...
 export const findDomainByNameOrToken = async (
   domainOrToken: string,
 ): Promise<Domain | undefined> => {
   const tokenName = normalizeDomainOrToken(domainOrToken);
   const domainName = normalizeDomainName(domainOrToken);
 
-  let domain =
-    (await Domain.findByNode(tokenName, undefined, true)) ||
-    (await Domain.findOnChainNoSafe(tokenName));
+  const [domain, znsDomain] = await domainsBatchLoader.loadMany([
```
tunguyennnnn, line 35: It's good start, I won't block PR for this comment but It's better if we like a DomainRepository with API of findMany wrapping the batcher. Service should not care how data is being fetched.

src/api.ts
```diff
@@ -65,56 +65,61 @@ if (env.APPLICATION.BUGSNAG_API_KEY) {
...
+  api.get('/api-docs/swagger.json', (_req: Request, res: Response) =>
+    res.json(swaggerSpec),
+  );
 
-api.get('/api-docs', (_req: Request, res: Response) => {
-  res.redirect('https://docs.unstoppabledomains.com/openapi/resolution/');
-});
+  api.get('/api-docs', (_req: Request, res: Response) => {
+    res.redirect('https://docs.unstoppabledomains.com/openapi/resolution/');
+  });
```
tirish, line 124: Minor: can keep this part out of the conditional

src/api.ts
```diff
@@ -65,56 +65,61 @@ if (env.APPLICATION.BUGSNAG_API_KEY) {
...
-swaggerSpec.paths['/domains'].get.parameters[0].style = 'deepObject';
+  // There's no way to set a custom attribute for a specific parameter in routing-controllers-openapi
+  // We could add a custom decorator for attributes in the future if we need to set more attributes
+  // But it's easier to just hard-code it for now
+  swaggerSpec.paths['/domains'].get.parameters[0].style = 'deepObject';
 
-api.get('/_status', (_: Request, res: Response) => {
-  res.json({
-    typeOrmCacheStats: InMemoryCache.getStatistics(),
+  api.get('/_status', (_: Request, res: Response) => {
```
tirish, line 112: Should this be conditional?

src/env.ts
```diff
@@ -74,7 +74,8 @@ type RunningMode =
   | 'METADATA_API'
   | 'ETH_WORKER'
   | 'MATIC_WORKER'
-  | 'ZIL_WORKER';
+  | 'ZIL_WORKER'
+  | 'API_DOCS';
```
tirish, line 78: Minor: the `_DOCS` part is a bit misleading now that doc hosting is removed, maybe `_SCHEMA` or `_SPEC`? 🤷

src/services/MetadataService.ts
```diff
@@ -299,7 +299,7 @@ export class MetadataService {
...
+        value: domain.children?.length || 0,
```
tirish, line 302: I see you added `children` to the `relations`, should we load the hydrated subdomain list every time?

src/services/DomainService.ts
tunguyennnnn: yes `setImmediate` will call in the next tick, but i was wondering is it enough time to accumulate domains from other requests. I was thinking of setTimeout(0) but an actually time out of 100 ms for example. 
if you look at this. it's actually the promise queue in the middle will be exhausted right after the any of the `timeout`, `callback`, `setimmediate` get called
sorry about the picture 🥲 but it's the only one I found where there's an arrow pointing to middle which is the promise queue.
Screenshot 2023-03-22 at 3 17 05 PM

src/services/DomainService.ts
DancingAxolotl: Yes, It's probably a good idea to use this throughout the repository if it gives significant performance improvements.

src/services/MetadataService.ts
DancingAxolotl: It's only loaded by the `DomainsService` which is only used by the metadata API.

src/controllers/StatusController.ts
DancingAxolotl: We don't have much traffic here, but its latency is very bad and it drives our overall p99 up. We don't really need up to date block numbers here, we just need to know if the worker is stuck.

We can exclude this change for now and create a separate ticket to revisit it later.

src/services/DomainService.ts
DancingAxolotl: I don't like having a non-zero timeout because it sets a minimum latency for our requests. So when we don't have spikes of traffic, one request would have to wait for 100ms to complete just because it's the minimum timeout.
The idea is that it will wait for all pending requests to come in and execute everything that we have collected so far without adding any extra latency. 

I'm pretty sure the default implementation of `dataloader` would achieve it too, but it uses `process.nextTick` under the hood and it may have some [unexpected side-effects](https://nodejs.org/ru/docs/guides/event-loop-timers-and-nexttick#why-would-that-be-allowed).

tunguyennnnn: >So when we don't have spikes of traffic, one request would have to wait for 100ms to complete just because it's the minimum timeout
make sense but I am curious how pending requests will be pushed to the batch queue. can you let me know the query log when you run it on staging please 🙏

src/services/MetadataService.ts
tirish: Do we really need a fully hydrated list, though?

src/controllers/StatusController.ts
DancingAxolotl: Discussed it with @sammyluo . We can keep this approach for now as there aren't any updates to the worker status table coming up. I created a linear ticket to reconsider this approach in the future.
https://linear.app/unstoppable-domains/issue/DT-1168/resolution-service-reconsider-status-endpoint

src/services/MetadataService.ts
DancingAxolotl: You are right, we don't really need it here. But it's better than doing a separate `count` query like we did before. It just adds a join to the existing query which doesn't add much latency.

src/services/DomainService.ts
DancingAxolotl: I changed the `/domains/:domainName` endpoint to also use the `DomainService` with batched queries too. I think it should give a similar performance improvement.

I wanted to try adding batching to the repository itself, but one big issue that I ran into is that most of our queries are not easy to batch. 
Even for simple queries like `get domain by name` there could be different parameters (`relations` is the most common one) that prevent these queries from being batched easily. So we either need to have generic batch queries which fetch everything (similar to what we have here) or setup separate loaders for each possible query option.

So it's not going to be easy to set this up.

src/services/DomainService.ts
DancingAxolotl: I added a simple stats library to count average, min and max batch size. It will be available in the `_status` endpoint.
We can use it for testing and remove it before pushing to prod.

tunguyennnnn (approved):

src/controllers/StatusController.ts
DancingAxolotl: I found another problem with this approach, it breaks the ZIL worker status because it turns out the status is not updated on each worker cycle. So for ZIL which has very rare transactions, the status updates rarely too and it's always out of sync. 

I moved these changes to a different PR so we can address them later #328

package.json
DancingAxolotl: I did some experimenting. And for some reason `batchloader` is more stable than `dataloader`. Ultimately it provides similar performance improvements, but `batchloader` is ~20% better.
So I'm not really sure which one is better.

DancingAxolotl: I ended up going with the batch loader library as it is easier to use and more stable based on the load testing. The way it's implemented it will be fairly easy to replace the library in the future if we need to.

src/api.ts
```diff
@@ -65,56 +66,62 @@ if (env.APPLICATION.BUGSNAG_API_KEY) {
...
+}
+
+api.get('/api-docs', (_req: Request, res: Response) => {
+  res.redirect('https://docs.unstoppabledomains.com/openapi/resolution/');
+});
 
 api.get('/_status', (_: Request, res: Response) => {
   res.json({
     typeOrmCacheStats: InMemoryCache.getStatistics(),
+    batchingStats: batchingStats,
```
tunguyennnnn, line 125: nit: 
```suggestion
    batchingStats,
```

src/models/Domain.ts
```diff
@@ -137,29 +132,19 @@ export default class Domain extends Model {
...
+    relations: string[] = ['resolutions', 'parent'],
   ): Promise<Domain[]> {
     if (!nodes.length) {
       return [];
     }
 
     return repository.find({
       where: { node: In(nodes) },
-      relations: ['resolutions', 'parent'],
+      relations: relations,
```
tunguyennnnn, line 147:
```suggestion
      relations,
```

src/services/DomainService.ts
```diff
@@ -1,44 +1,48 @@
...
+const domainsBatchLoader = buildBatchLoader(
+  (keys: readonly string[]) =>
+    Domain.findAllByNodes(keys as string[], undefined, true, [
+      'resolutions',
+      'reverseResolutions',
+      'parent',
+      'children',
+    ]),
+  (domain) => domain.node,
+  {
```
tunguyennnnn, line 20: maybe later if we want to optimize more, batch load should have a cache interface we can use

tunguyennnnn (approved):
very nice!!!

PR #317, opened 2023-03-22
Fix profile query params

DancingAxolotl:
## Background

Minor fix for profile PFPs

## Changes

- added param to fetch profile only


tirish (approved):

qrtp (approved):
Just want to validate behavior in staging with this domain, which includes an off-chain image:

https://metadata.ud-staging.com/image-src/aaron.x.svg

Should look like this:
image

PR #323, opened 2023-03-24
anhan: undo changes from dth-7

anhan32:
undo code changes from https://linear.app/unstoppable-domains/issue/DTH-7/bug-case-sensitive-get-records-for-owner-addresses-args

The lowercase query was causing db latency

### DIFF
--- a/src/controllers/DomainsController.ts
+++ b/src/controllers/DomainsController.ts
@@ -146,13 +146,12 @@ export class DomainsController {
     if (query.resolution) {
       const resolutionKeys = Object.keys(query.resolution);
       for (let i = 0; i < resolutionKeys.length; i++) {
-        const key = Object.keys(query.resolution)[i];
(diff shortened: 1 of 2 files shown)


src/controllers/DomainsController.ts
```diff
@@ -146,13 +146,13 @@ export class DomainsController {
     if (query.resolution) {
       const resolutionKeys = Object.keys(query.resolution);
       for (let i = 0; i < resolutionKeys.length; i++) {
-        // key is validated within DomainsListQuery validation to be one of the expected resolver keys
-        // the value is validated to be a string
-        const key = resolutionKeys[i];
+        const key = Object.keys(query.resolution)[i];
```
tirish, line 149:
```suggestion
        const key = resolutionKeys[i];
```

tirish (approved):

PR #333, opened 2023-04-07
[DT-1195] fix: Extract batchloader library, re-write cache loader

DancingAxolotl:
## Background

 - Adding the batchloader library introduced a memory leak, also the library is not actively supported. We need to extract the library into our repo to make it easier for us to make changes & support it.

## Changes

 - Removed batchloader dependency, extracted into separate module in `utils`
 - Re-wrote cache loader using `node-cache`
 - Updated `buildBatchLoader` function and its usage

## TODO

 - [x] batchloader is licensed under the MIT license so we should be okay to use it this way. However, we need to properly credit the author. For now I added a readme in the batch loader folder, we need to make sure that this is enough.

## Code Review

- [x] **Implementation** - satisfied acceptance criteria
- [x] **Testing** - unit tests pulled from the original repo, verified locally that memory consumption is greatly reduced
- [x] **Performance** - optimized expensive algorithms & database queries

## Confirmation


src/env.ts
```diff
@@ -263,6 +263,10 @@ export const env = {
               process.env.QUERY_BATCHING_CACHING_ENABLED,
               true,
             ),
+      CACHE_TTL_MS: parseNumberFromEnv(
+        process.env.QUERY_BATCHING_CACHE_TTL_MS,
+        600000, // 10 minutes
```
tunguyennnnn, line 268: 🤔 I am bit concerned about this. this means if we have 500 distinct req/s per instance it would be, it would be 600 x 500 = 300k records stored in memory per time. not sure how much is in in bytes but it sounds a lot. I think this scenario is possible when some partners mirror records from us.

src/utils/BatchLoader/batchLoader.ts
```diff
@@ -0,0 +1,121 @@
+import { IBatchLoader } from './types';
+import { PossiblePromise } from '../../types/common';
+
+export type BatchLoadFn<Key, Value> = (
+  keys: Key[],
+) => Value[] | Promise<Value[]>;
+
+export type KeyToUniqueId<Key> = (key: Key) => string;
+
+const sleep = (ms: number): Promise<void> =>
```
tunguyennnnn, line 10: should put this util into a util file. maybe promiseUtil?

src/utils/BatchLoader/batchLoader.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export class BatchLoader<Key, Value> implements IBatchLoader<Key, Value> {
+  protected queuedKeys: Key[] = [];
+  protected batchPromise: Promise<Value[]> | null = null;
+
+  constructor(
+    protected batchFn: BatchLoadFn<Key, Value>,
+    protected keyToUniqueId: KeyToUniqueId<Key> | null,
+    protected batchDelay = 0,
+    protected batchSize = Number.MAX_SAFE_INTEGER,
```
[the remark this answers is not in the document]
DancingAxolotl: Actually, we can take the `env.BATCHING` value here directly. I didn't really touch the `batchLoader` implementation. We can probably clean this up a little.

src/env.ts
DancingAxolotl: That's a good concern, thanks for bringing it up. From `node-cache` docs, there is no limit to the cache size. So it will just consume the whole heap and run out of memory. There is no way to limit it out of the box. We can limit the total number of cache keys in node-cache, but we would have to manage keys ourselves since it does not delete keys automatically when it hits the limit. https://github.com/node-cache/node-cache#options - the `maxKeys` option

At the same time, I don't think we will ever hit 500 distinct req/s the way we have it configured currently (at least because of concurrency limits on GCP). 

I think for now we can just keep it as 10 minutes and keep an eye on cache stats in `_status`.

PR #315, opened 2023-03-21
[DTH-17] chore: Implement OpenSea asset request without SDK

tirish:
## Background

For various reasons, we want to remove the OpenSea SDK. Since we are only using it for a single request (get single asset) just implement that directly on our own.

## Changes

- Remove `opensea-js`
- Implement a simple OpenSea API Client for the "get single asset" route
- Included retry logic to mimic logic from the OpenSea SDK

## To Do

- [x] confirm testing on testnet
- [x] confirm testing on mainnet

## Additional Deployment and/or Rollback Steps

n/a

## Code Review

This Pull Request was thoroughly reviewed and meets all Code Review Standards pertaining to:

- [x] **Implementation** - satisfied acceptance criteria
- [x] **Communication** - notified engineers/stakeholders about impactful changes
- [x] **Error handling** - handled, logged & alerted on errors
- [ ] **Documentation** - wrote readable code & commits, documented unintuitive code
- [x] **Testing** - unit & E2E test coverage, tested in staging, DB migrations backwards compatible
- [ ] **Security** - sanitized inputs, vetted dependencies, validated/secured API requests, no committed secrets
- [ ] **Performance** - optimized expensive algorithms & database queries

Please select all **applied** standards relevant to this PR.
## Confirmation

- [x] **Assignee Confirmation** - I (the author) have filled out the template above
- [ ] **Reviewer Confirmation** - I (a/the reviewer) confirm 1) the author has filled out the template and checked the box that says **Assignee Confirmation** 2) I reviewed the code and selected all applicable items in the code review section.

PR #336, opened 2023-04-10
[DT-1186] fix: Replace opensea SDK with API, attempt 2

DancingAxolotl:
## Background

Previous attempt in #315 caused some increased latency. Updated implementation to match SDK behaviour.

## Changes

 - Added back retries for failed requests similar to SDK implementation
 - Added 500ms timeout for API requests based on average latency from https://status.opensea.io/

## Code Review

## Confirmation


src/services/OpenSeaService.ts
```diff
@@ -21,26 +16,46 @@ export class OpenSeaService {
     logger.warn(`${msg}: ${res.status} | ${content}`);
   }
 
+  private async fetchRetry(
+    url: string,
+    opts?: RequestInit,
+    retries = 1,
+    pause = 1000,
+  ): Promise<Response> {
+    const response = await fetch(url, opts);
```
tunguyennnnn, line 25: 🤔 why we need to retry? if request fail the first time we should just fail it, and let the clients to request again

DancingAxolotl: It's a difficult question.

There are a few reasons in favour of retrying:
 - The SDK has this. Since we don't know why SDK and API behave differently I wanted to match the SDK as closely as possible.
 - The client may not get a chance to request again. The error that we throw is not blocking, so we will cache the default image and we will not request it again. 

At the same time, retrying will add latency. And if the first request failed for some reason, there's a good chance it will fail the second time.

Since we already tried it without retries, I'm thinking that we should try with retries for now. And we can remove them if the latency is still bad.

tunguyennnnn (approved):

PR #338, opened 2023-04-12
Fix: Fix caching time unit

DancingAxolotl:
## Background

 - `node-cache` expects a ttl in seconds, while we specify it in milliseconds, so we ended up with 1000x cache ttl.

## Changes

- convert ttl units


src/utils/BatchLoader/cacheLoader.ts
```diff
@@ -18,7 +18,7 @@ export class CacheLoader<Value> implements IBatchLoader<string, Value> {
     if (value === undefined) {
       return this.loader.load(key).then((val) => {
         if (this.ttl) {
-          this.cache.set(key, val, this.ttl);
+          this.cache.set(key, val, this.ttl / 1000);
```
sammyluo, line 21: Would be great to leave a comment to indicate `ttl`'s unit

sammyluo (approved):

tunguyennnnn (approved):

PR #340, opened 2023-04-13
[DT-1233] feat: store domain records in a table instead of json blob

DancingAxolotl:
## Background
Looking up domains by records and accessing specific records is slow since they are stored in a json blob within postgres. This PR changes storage to a separate table in the DB.

## Changes
 - Added a new table for domain resolution records
 - Updated workers to save records to the new table
 - Updated API controllers to read from the new table
 - Some more changes to cleanup the code & fix unit tests in #353 

## Deployment and rollback steps
Deployment:
 1. Deploy #355 first
 2. Wait for the script to finish
 3. Rebase #340  onto master and deploy APIs

Rollback:
 1. Rollback code changes
 2. Run rollback migration to delete the new DB

## To Do
 - [x] Some more testing to make sure that all endpoints work + load testing on testnet

## Code Review

## Confirmation


src/models/DomainsResolutionRecords.ts
```diff
@@ -0,0 +1,67 @@
...
+  @Column({
+    name: 'value',
+    comment: 'resolution record value',
+    unique: false,
+    nullable: true,
+  })
+  @IsString()
+  // @ValidateWith<DomainsResolutionRecords>('validRecord', {
+  //   message: 'Invalid resolution record',
+  // })
```
DancingAxolotl, line 46: We have the option to validate records with regexes in the `resolver-keys.json` but as you would expect we have lots of records on-chain which don't match our regexes. Adding this validation will make these records invalid. This could be a good thing - avoids sending crypto to invalid addresses; and a bad thing - breaks a bunch of existing records. We should think about this.

src/workers/framework/WorkerRepository.ts
```diff
@@ -134,6 +136,19 @@ export class WorkerRepository implements IWorkerRepository {
...
+            if (
+              Object.prototype.hasOwnProperty.call(resolution.resolution, key)
+            ) {
+              const value = resolution.resolution[key];
+              dbResolution.setRecord(key, value);
+            }
+          }
+        } else {
+          dbResolution.records = [];
+        }
```
DancingAxolotl, line 150: Untested

DancingAxolotl (commented):
Left some notes inline for future implementation

src/models/DomainsResolutionRecords.ts
DancingAxolotl: For now I am removing this validation. After thinking about it, it introduces a lot more problems than it solves.

There is no way for us to control which records are placed on-chain, so there will always be invalid/broken records. Since the goal of the resolution service is to mirror the state of the blockchain, it has to return these records as is. Regardless if they match any of our conventional formats.

tunguyennnnn (changes requested):
migration will take very long and the code relying on the new table. 
I think we need to make it into either:
* deploying a separate PR that populate the new table, and worker starts to write both to new table and the old json blob. -> backfill the table -> deploy the code that reads from the new table
OR 
* having bridge code, it should lookup into the new table and fallback to old json blob if no record is found.

src/controllers/DomainsController.ts
```diff
@@ -401,8 +412,17 @@ export class DomainsController {
       const domain = allDomains.find((d) => d.name === domainName);
 
       if (domain) {
-        const { resolution } = getDomainResolution(domain);
-        const records = query.key ? pick(resolution, query.key) : resolution;
+        const resolution = getDomainResolution(domain);
+        const records = (
+          query.key
+            ? resolution.records.filter((kv) => query.key == kv.key)
```
tunguyennnnn, line 418:
```suggestion
            ? resolution.records.filter((kv) => query.key === kv.key)
```

src/controllers/DomainsController.ts
```diff
@@ -81,7 +81,12 @@ export class DomainsController {
         registry: resolution.registry,
         reverse: domain.hasReverseResolution,
       };
-      response.records = resolution.resolution;
+      response.records = resolution.records.reduce((records, record) => {
```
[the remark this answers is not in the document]
DancingAxolotl: Yes, the value column could be empty. For some reason we have some empty resolutions in the DB already. And theoretically you can set a bunch of empty records in the smart contract. So just to clean up the response I'm not returning them in the final json.

DancingAxolotl:
I pulled some changes from #353 into this PR to have everything implementation-related here. So only unit test fixes are left in #353 

> migration will take very long and the code relying on the new table.
> I think we need to make it into either:
> deploying a separate PR that populate the new table, and worker starts to write both to new table and the old json blob. -> backfill the table -> deploy the code that reads from the new table

@tunguyennnnn Yes, that is the plan for now.

tunguyennnnn (approved):
LGTM!

DancingAxolotl:
Screenshot 2023-05-11 at 15 08 57

PR #388, opened 2023-08-02
[DT-1002] feat: framework changes for ENS support

DancingAxolotl:
## Background

Split out from #379, this PR contains the framework changes that were added for ENS support and some performance improvements to speed up initial sync for workers.

## Changes

 - Dependencies
   - upgraded uns dependency
   - added some ens dependencies
   - fixed controller tests that used uns sandbox
 - Blockchain configs
   - Moving from UNS-specific config to a generic EVM config for smart contracts
   - Renaming/moving files:
      - src/contracts -> src/blockchain/*
      - EthereumProvider -> src/blockchain/evm/EvmProvider.ts
      - Replace `getEthConfig` with `getEvmConfig` to load a generic set of ABIs
 - DB models
   - src/models/Domain.ts
     - Removed `nameMatchesNode` validation to support domains with unknown names
     - Fixed `setParent` to set the `parent_id` instead of `parent` because `typeorm` gets confused
     - Fixed relations in `findAllByNodes`
   - src/models/DomainsResolution.ts
     - Removed validation on `resolver` to allow null resolver address
   - src/models/NamehashLookup.ts
     - Fixed hash calculation
   - src/models/WorkerStatus.ts
     - Allow `location` to be a generic string instead of `Blockchain`
 - Utils
   - Some new generic utils (`isEmpty`, `mergeDeep`)
   - Fixed testing utils for upgraded uns sandbox
 - Workers
   - Moved uns workers and tests into a separate folder
   - Fixes for framework and config changes
 - Worker framework
   - Base worker
     - Add an async `setup` method to base worker to allow for async initialisation steps
     - Commit to DB on every sync iteration in `run` instead of once after everything is synced
     - Handle event loading errors by re-trying with a smaller block range
   - Worker repository
     - Interface functions are now non-blocking
     - Changes are saved to the DB upon calling `commit`
     - Changes are stored in an in-memory cache before committing to DB
     - Added a batch loader for reading `Domain` data
     - Save events data in chunks because typeorm can't handle too many query parameters (>10k)

## Code Review

## Confirmation


src/workers/framework/WorkerRepository.ts
```diff
@@ -58,27 +105,185 @@ export class WorkerRepository implements IWorkerRepository {
...
+    return Object.fromEntries(namehashes.map((n) => [n.hash, n.value]));
+  }
+
+  private async _saveDomains(domain: Domain | Domain[]): Promise<void> {
     // find all nodes in DB
     const domains = convertToArray(domain);
     const domainsToSave: DbDomain[] = [];
+    const lookupValues: string[] = [];
+    const dbDomains = this.preloadDomains(
+      domains.reduce((a, d) => {
```
tunguyennnnn, line 236: I'd elaborate naming here so it's easy to read:
```
domains.reduce((domainList, domain) => {
```

src/blockchain/evm/EvmContracts.ts
```diff
@@ -0,0 +1,78 @@
+import type { JsonFragment } from '@ethersproject/abi';
+import { Contract } from 'ethers';
+import { UnsContractName } from 'uns';
+import UnsNetworkConfig from 'uns/uns-config.json';
+
+import { StaticJsonRpcProvider } from './EvmProvider';
+
+type Abi = ReadonlyArray<JsonFragment>;
+
+const getAbi = async <T extends string>(name: T): Promise<Abi> => {
```
tunguyennnnn, line 10: maybe just throw instead of catch it? it's critical to have the correct file name?

src/blockchain/evm/EvmContracts.ts
```diff
@@ -0,0 +1,78 @@
...
+  try {
+    const lib = await import(`uns/artifacts/abi/${name}.json`);
+    return lib.default;
+  } catch (error) {
+    return [];
+  }
+};
+
+const AbiMapGenerator = <T extends string>(contracts: T[]) => {
+  return Promise.all(
```
tunguyennnnn, line 20: 😄 i'd elaborate the naming instead of a, c, v

src/blockchain/evm/EvmContracts.ts
```diff
@@ -0,0 +1,78 @@
...
+  networkId: string,
+  abiMap: Record<string, Abi>,
+  provider: StaticJsonRpcProvider,
+  networks: { [key: string]: { contracts: any } },
+): CryptoConfig => {
+  const cryptoConfig = {} as CryptoConfig;
+  if (Object.keys(networks).includes(networkId)) {
+    const { contracts } = networks[networkId];
+
+    Object.keys(contracts).forEach((key) => {
```
tunguyennnnn, line 48: we should use for in loop for easier to read
and use `contractName` or some better name to say what it is
```suggestion
    for (const contractName in contracts) {
     ...
    }
```

src/models/Domain.ts
```diff
@@ -89,25 +86,29 @@ export default class Domain extends FastlyCachableModel {
...
-      this.parent = await this.setParent();
+    if (!this.parentId) {
+      this.parentId = (await this.setParent())?.id || null;
     }
   }
 
   private async setParent(): Promise<Domain | null> {
-    const splittedName = this.getSplittedName();
-    let parentName = splittedName[splittedName.length - 1];
+    if (this.name.indexOf('.') > 0) {
```
tunguyennnnn, line 95: lets make the function more flat:

```
```suggestion
    if (this.name.indexOf('.') === 0) {
      return null;
    }
    ...
```

tunguyennnnn: worth to make this a util in utils/domain?
something like getParentDomain?

src/models/WorkerStatus.ts
```diff
@@ -46,31 +45,31 @@ export default class WorkerStatus extends Model {
   }
 
   static async latestMirroredBlockForWorker(
-    location: Blockchain,
+    location: string,
```
tunguyennnnn, line 48: 🤔 what the reason of changing to string?

src/utils/common.ts
```diff
@@ -11,3 +11,30 @@ export function chunk<T>(arr: T[], chunkSize: number): T[][] {
 
 // adding a timeout to help mitigate against long-running external dependencies
 export const PROFILE_FETCH_TIMEOUT_MS = 1000;
+
+export const isEmpty = (obj: any) =>
```
tunguyennnnn, line 15: can you use 'lodash' isEmpty instead?

src/utils/common.ts
```diff
@@ -11,3 +11,30 @@ export function chunk<T>(arr: T[], chunkSize: number): T[][] {
...
+  [Object, Array].includes((obj || {}).constructor) &&
+  !Object.entries(obj || {}).length;
+
+export function isObject(item: any) {
```
tunguyennnnn, line 19: lodash has this as well

src/utils/common.ts
```diff
@@ -11,3 +11,30 @@ export function chunk<T>(arr: T[], chunkSize: number): T[][] {
...
+  return item && typeof item === 'object' && !Array.isArray(item);
+}
+
+export function mergeDeep(target: any, source?: any) {
```
tunguyennnnn, line 23: lodash merge function satisfy what you need? See the result of what it does
Screenshot 2023-08-02 at 2 19 44 PM

tunguyennnnn (commented):
Overall looks good 🥇 , some small comments and suggestion for cleanup

tunguyennnnn (approved):
LGTM, Comments can be addressed in later PR to not block deployment

src/workers/framework/WorkerRepository.ts
DancingAxolotl: I simplified this to avoid using reduce. Should be easier to read now

src/blockchain/evm/EvmContracts.ts
DancingAxolotl: We have some contracts missing in `uns` (ProxyAdmin, WhitelistedMinter, TwitterValidationOperator). So even if we throw here we would have to catch and ignore the error. I added a log here so we don't miss it, but I don't think it's worth throwing.

src/models/WorkerStatus.ts
DancingAxolotl: Because it's not just the blockchain now. We have two workers on ETH so we need a different value to tell workers apart. For now the ENS worker uses `ENS` and UNS uses `MATIC` and `ETH`, but we can probably change it to something like `ENS-ETH` and `UNS-MATIC` later.

src/models/Domain.ts
DancingAxolotl: Refactored and added some tests.
I don't think it's worth it to make this a util since in general you would just use the `parent` relation to find the parent. This function is only necessary specifically to fill in the parent id before saving the entity.

PR #400, opened 2023-08-16
[DT 1511] Add owner domains api

tunguyennnnn:
## Background

Introduce this API to support get domains by ownerAddress for muti-naming services instead of relying on legacy List domain API

## Changes

* add batchloader that return multi records for resolution
* add /owners/:address/domains API
* Update doc with example

## Code Review

## Confirmation


src/services/DomainService.ts
```diff
@@ -64,3 +64,32 @@ export const findDomainByNameOrToken = async (
...
+
+  const qb = Domain.createQueryBuilder('domain');
+  qb.leftJoinAndSelect(
+    'domain.resolutions',
+    'resolution',
+    `resolution.domain > :resDomainId`,
+    { resDomainId: afterId },
+  );
+
+  qb.leftJoinAndSelect('resolution.records', 'records');
```
DancingAxolotl, line 85: Do we need to join in `domain.metadata` too?

DancingAxolotl: Also, maybe `parent`?
Or make these columns optional?

src/services/DomainService.ts
```diff
@@ -64,3 +64,32 @@ export const findDomainByNameOrToken = async (
...
+
+  qb.take(limit);
+
+  qb.where(`resolution.owner_address = :address`, {
+    address,
+  });
+
+  const domains = await qb.getMany();
```
DancingAxolotl, line 93: This probably is not an issue now. But I wonder how it would behave with multiple resolutions per domain? Would we get duplicate domains?

src/services/nameing-services/EnsNamingService.ts
```diff
@@ -36,6 +46,10 @@ export class EnsNamingService extends NamingService {
     });
   }
 
+  /////////////////////////
+  //// Validators /////////
+  ////////////////////////
```
DancingAxolotl, line 51: TIL: You can define collapsible regions in VS Code. I was looking for something like this too.
https://burkeholland.gitbook.io/vs-code-can-do-that/exercise-3-navigation-and-refactoring/folding-sections#folding-by-region

src/services/nameing-services/EnsNamingService.ts
```diff
@@ -71,6 +85,37 @@ export class EnsNamingService extends NamingService {
...
+          );
+        });
+
+        if (isMatched) {
+          return nameRecord;
+        }
+      }
+    }
+
+    // TODO: look up in namehash_lookup table for matching name
```
DancingAxolotl, line 114: We shouldn't need to look up `namehash_lookup` in the controllers. 
The workers already do the lookup when saving domains. So if it's in the lookup table, it will be set in the DB.

DancingAxolotl (approved):
Nice work @tunguyennnnn 🔥 
Overall LGTM, just some minor inline comments

src/services/DomainService.ts
tunguyennnnn: > Do we need to join in domain.metadata too?
yea I updated it in in https://github.com/unstoppabledomains/resolution-service/pull/391, it's only needed for ENS

tunguyennnnn: > Also, maybe parent?
Or make these columns optional?

can you explain this?

src/services/nameing-services/EnsNamingService.ts
tunguyennnnn: ah, this is cool! let me try that

src/services/DomainService.ts
DancingAxolotl: I was just wondering about any relations that we may need here. We can add them later if we need to.

src/services/DomainService.ts
tunguyennnnn: it will return 1 domain with multiple resolutions. 
For UNS it will just return the resolution of the latest resolution

PR #411, opened 2023-08-23
[DT 1533] setup pubsub producer

tunguyennnnn:
## Code Review

## Confirmation


src/pubsub/PubSubSubscriber.ts
```diff
@@ -0,0 +1,76 @@
...
+  };
+
+  runListener = async (): Promise<void> => {
+    if (!(this.subscription instanceof Subscription)) {
+      throw new Error('Subscription is not an allow to listen to messages');
+    }
+
+    this.subscription.on('message', async (message: SubscriptionMessage) => {
+      try {
+        await this.handleMessage(message);
```
DancingAxolotl, line 63: I think it would be nice if the base subscriber could do some preliminary parsing for the message. I think it comes in as a raw buffer. So we can at least parse it into JSON.

But on top of that, since we define the message schema, we can probably let the base subscriber figure out specific event types. It will allow us to have specific handlers for each event instead of one big "catch-all"  `handleMessage(message)`

DancingAxolotl: I think it would be awesome if the final subscriber class could look like some of [ENS event handlers](https://github.com/unstoppabledomains/resolution-service/blob/master/src/workers/eth/ens/EnsController.ts) for example. 

So the implementation can just focus on writing the handlers (maybe even with typed params) and the utility can handle the rest within our agreed schemas and tools.

tunguyennnnn: > I think it would be nice if the base subscriber could do some preliminary parsing for the message. I think it comes in as a raw buffer. So we can at least parse it into JSON.
> 
> But on top of that, since we define the message schema, we can probably let the base subscriber figure out specific event types. It will allow us to have specific handlers for each event instead of one big "catch-all" `handleMessage(message)`

yea i would let the `handleMessage` itself to parse and validate the messages, we might not have all pubsub topic publishing json, it could be later decode protobuf instead.

tunguyennnnn: > I think it would be awesome if the final subscriber class could look like some of [ENS event handlers](https://github.com/unstoppabledomains/resolution-service/blob/96ae7920ee586ea1b416bf7772a29120a1b2af2c/src/workers/eth/ens/EnsController.ts#L42) for example.
> 
> So the implementation can just focus on writing the handlers (maybe even with typed params) and the utility can handle the rest within our agreed schemas and tools.

👍 , yea we can do that in the implementation. this PR is just the base to hook things up. Also we should have Service files to handle depending on the naming services

DancingAxolotl: > yea i would let the handleMessage itself to parse and validate the messages, we might not have all pubsub topic publishing json, it could be later decode protobuf instead.

Then it's even better to parse it in the base utility. This way we can switch to protobuf in one place and everyone will get the benefit without duplicating code. And even at this point I can see every `handleMessage` implementation being the same:
```
handleMessage(message) {
  try {
    const data = JSON.parse(message.data.toString());
    switch (data.topic) {
      case "...": {
        ...
      }
      case "...": {
        ...
      }
      case "...": {
        ...
      }
      default: {
        ...
      }
    }
  } catch {
    ...
  }
}
```

> 👍 , yea we can do that in the implementation. this PR is just the base to hook things up. 

In that case the base utility should support it. But yeah, from the task it sounds like the goal is to just setup PS for now, right? We can improve the boilerplate code in the future

tunguyennnnn: > In that case the base utility should support it. But yeah, from the task it sounds like the goal is to just setup PS for now, right? We can improve the boilerplate code in the future

yep

tunguyennnnn: >Then it's even better to parse it in the base utility. This way we can switch to protobuf in one place and everyone will get the benefit without duplicating code. And even at this point I can see every handleMessage implementation being the same:

how about having subclass implement a `parse` function? the baseclass can just implement json parsing but it should not know about all the topics

package.json
```diff
@@ -20,8 +20,14 @@
...
-    "start:dev": "dotenv -e ./local.dev.env -- ts-node src/index.ts",
-    "start:dev:watch": "dotenv -e ./local.dev.env -- nodemon --exec ts-node src/index.ts",
+    "start:api:dev": "SERVICE=API_SERVICE dotenv -e ./local.dev.env -- ts-node src/index.ts",
+    "start:api:dev:watch": "SERVICE=API_SERVICE dotenv -e ./local.dev.env -- nodemon --exec ts-node src/index.ts",
+    "pubsub:containers:dev": "docker-compose -f dev/docker-compose.yml up && yarn dev:pubsub:setup",
+    "pubsub:containers:dev:down": "docker-compose -f dev/docker-compose.yml down",
+    "start:subscriber:dev": "dotenv -e ./local.dev.env -- ts-node src/domainSubscriberService.ts",
+    "start:subscriber:dev:watch": "dotenv -e ./local.dev.env -- nodemon --exec ts-node src/domainSubscriberService.ts",
+    "start:worker:dev": "SERVICE=WORKER_SERVICE dotenv -e ./local.dev.env -- ts-node src/index.ts",
+    "start:worker:dev:watch": "SERVICE=WORKER_SERVICE dotenv -e ./local.dev.env -- nodemon --exec ts-node src/index.ts",
```
DancingAxolotl, line 30: Do you mind leaving a quick note in the readme about new commands to start the new dev servers and the new service env. var?

src/index.ts
```diff
@@ -30,6 +30,9 @@ void connect().then(async () => {
   //   }
   //   logger.info('Db snapshot loaded');
   // }
+
+  const isDevelopmentMode: boolean = process.env.NODE_ENV === 'dev';
```
DancingAxolotl, line 34: Should we move this to `env.ts`?

dev/Dockerfile.SetupTopics
```diff
@@ -0,0 +1,5 @@
+FROM node:14.21.2
+
+WORKDIR /usr/src/app
+
+ENTRYPOINT ["yarn", "dev:pubsub:setup"]
```
DancingAxolotl, line 5: This probably needs an `ADD` command to copy over the RS files to the contaner
And maybe some dependencies

See https://github.com/unstoppabledomains/resolution-service/blob/master/Dockerfile for example

src/pubsub/PubSubSubscriber.ts
DancingAxolotl: Yeah, that would work too.

We can also pass some topic info to the base class to let it know what to parse without defining the topics explicitly, or have the base class collect topics from subclass attributes. But we can add all of this later on. I think this should be fine for now.

DancingAxolotl (approved):
Nice start 🚀 
Overall LGTM. Just a few minor comments, but nothing blocking

dev/Dockerfile.SetupTopics
tunguyennnnn: yea i would, but because the docker-compose already point the whole volumes to local so it's not necessary

PR #425, opened 2023-09-15
[Hotfix]: Revert ens separate process

DancingAxolotl:
## Background

Now that the ENS sync is finished we can revert running it in a separate process.

## Changes

Reverts changes from #408


bcarl-ud (approved):
👍

tunguyennnnn (approved):

PR #431, opened 2023-09-20
Refactor of general image, with design updates, removed qr code param

bcarl-ud:
-New design updates for PFP
-Refactored and removed the QR Code param


bcarl-ud:
Closing in favor of using xhtml namespace and css styling for domain label.

PR #440, opened 2023-09-27
[DT-1556] feat: pubsub producer consumer

DancingAxolotl:
## Background

Adding producers and consumers for pubsub

## Changes

 - Updated pubsub utils `src/pubsub`
 - Added producer and consumer for domain events: 
 - `src/workers/framework/WorkerProducer.ts` - runs with worker repository (if `env.PUBSUB.ENABLED == true`)
 - `src/workers/framework/WorkerConsumer.ts` - runs in `index.ts` if `RUNNING_MODE` includes `PUBSUB_SUBSCRIBER`

## Code Review

## Confirmation


tunguyennnnn (approved):

src/workers/framework/WorkerRepository.ts
```diff
@@ -40,6 +39,8 @@ import {
   WorkerEvent,
   merge,
 } from './Types';
+import { WorkerProducer } from './WorkerProducer';
+import { getLocation } from './common';
```
DancingAxolotl, line 43: We should reuse other functions in `./common`

src/pubsub/types.ts
```diff
@@ -0,0 +1,50 @@
...
+  IsOptional,
+  IsString,
+  ValidateNested,
+} from 'class-validator';
+
+import { Attributes } from '../types/common';
+
+export class PubSubError extends Error {}
+
+export class PubSubMessage<T extends object> {
```
DancingAxolotl, line 12: I think this will also need `zod` validation, so that nested validation works properly. Or just remove validation from PubSubMessage entirely and validate just the internal data.

src/workers/framework/Types.ts
```diff
@@ -219,3 +227,28 @@ export function merge<T extends Domain | Resolution>(
...
 }
+
+export enum WorkerEventTypes {
+  Domain = 'Domain',
+  Resolution = 'Resolution',
+  SetReverseResolution = 'SetReverseResolution',
+  UnsetReverseResolution = 'UnsetReverseResolution',
+}
+
+export class DomainPubSubAction<
```
DancingAxolotl, line 238: How is this different from `DomainAction`? Shouldn't it match zod/avro schema?

src/workers/framework/WorkerProducer.ts
```diff
@@ -0,0 +1,313 @@
...
+              ? {
+                  node: rr.event.node,
+                }
+              : undefined,
+          }),
+      ),
+    );
+
+    await this.publishBatchEvents(
+      this.unsetReverseResolutionCache.map(
```
DancingAxolotl, line 290: We should probably keep `set` and `unset` events together for RR to ensure the order of published events.
For example, if we have a batch of events that includes `unsetRR1, setRR1, unsetRR2, setRR2` if we drop group them by action we will get: `setRR1, setRR2, unsetRR1, unsetRR2` Which results in wrong RR.

Ideally we should just be smart about merging RR events to keep a flat structure, but that logic is tricky. So for now we can just keep them in one list.

src/workers/framework/Types.ts
tunguyennnnn: it requires event to have object such with keys in avro schema, for example:
```
{
  ...
  event: {
     'com.ud.resolution.Domain': {
        node: ....,
        ...
     }
  }
```
we could return the transformation from Domain Action as well.

src/pubsub/types.ts
tunguyennnnn: do you mean validate to the level of node, type, etc..?
unless we want to validate like ethereum address format (ex: regex), the  Avro calls will validate for us in here
https://github.com/unstoppabledomains/resolution-service/pull/440/files#diff-2d9b3725d604ceb997201e4908f83534a8f376edb142106b7e4240f07bc57b74R112
it will reject, but we can log the message there.

PR #450, opened 2023-10-03
[DT 1567] Decoding ens records

tunguyennnnn:
## Code Review

## Confirmation


src/services/naming-services/EnsNamingService.ts
```diff
@@ -74,9 +97,94 @@ export class EnsNamingService extends NamingService {
...
 
+  //////////////////////
+  // decode records   //
+  //////////////////////
+
+  private decodeContentHash(contentHash: string): string {
+    try {
+      const codec = EnsContentHash.getCodec(contentHash);
+      let decodedValue = EnsContentHash.decode(contentHash);
+      if (codec === 'ipfs') {
```
sammyluo, line 108: Should we handle ipns as well?

src/services/naming-services/EnsNamingService.ts
```diff
@@ -74,9 +97,94 @@ export class EnsNamingService extends NamingService {
...
+      return `/${codec}/${decodedValue}`;
+    } catch (error) {
+      return contentHash;
+    }
+  }
+
+  private decodeAddr(key: string, value: string): string {
+    try {
+      const keyComponents = key.split('.');
+      if (keyComponents.length !== 2) {
```
sammyluo, line 121: it's probably safer to check the record starts with `addr` as well

src/services/naming-services/EnsNamingService.ts
```diff
@@ -74,9 +97,94 @@ export class EnsNamingService extends NamingService {
...
+        return value;
+      }
+
+      const coinType = parseInt(keyComponents[1], 10);
+      if (!_.isInteger(coinType)) {
+        return value;
+      }
+      const format = EnsAddressEncoder.formatsByCoinType[coinType];
+      const bufferValue = prepareHexValue(value);
+      return format.encoder(Buffer.from(bufferValue, 'hex'));
```
sammyluo, line 131: Should we check the existence of `format` first before using it?

src/services/naming-services/EnsNamingService.ts
```diff
@@ -74,9 +97,94 @@ export class EnsNamingService extends NamingService {
...
+        return {
+          decodedKey: FabricatedRecordsTypes.AbiUnknown,
+          decodedValue: value,
+        };
+      }
+      const bufferValue = prepareHexValue(value);
+      const jsonString = Buffer.from(bufferValue, 'hex').toString('utf8');
+      return {
+        decodedKey: FabricatedRecordsTypes.AbiJson,
+        decodedValue: JSON.parse(jsonString),
```
sammyluo, line 155: The return type of JSON.parse is no longer a string, right? So we should either update the return type of the function, or do
```suggestion
        decodedValue: JSON.stringify(JSON.parse(jsonString)),
```

src/services/naming-services/EnsNamingService.ts
```diff
@@ -74,9 +97,94 @@ export class EnsNamingService extends NamingService {
...
+  ): { decodedKey: string; decodedValue: string } {
+    try {
+      const keyComponents = key.split('.');
+      // ABI can have multiple formats, we only support plain JSON which is 1
+      // https://docs.ens.domains/ens-improvement-proposals/ensip-4-support-for-contract-abis
+      if (keyComponents.length !== 2 || keyComponents[1] !== '0x01') {
+        return {
+          decodedKey: FabricatedRecordsTypes.AbiUnknown,
+          decodedValue: value,
+        };
```
sammyluo, line 149: Nit: why not just throw an error here? The catch returns the same value anyway.

src/services/naming-services/EnsNamingService.ts
```diff
@@ -74,9 +97,94 @@ export class EnsNamingService extends NamingService {
...
+  ): Record<string, string> {
+    for (const [key, value] of Object.entries(records)) {
+      if (key === EnsRecordType.Contenthash) {
+        records[key] = this.decodeContentHash(value);
+      } else if (key.startsWith(EnsRecordType.Addr)) {
+        records[key] = this.decodeAddr(key, value);
+      } else if (key.startsWith(EnsRecordType.Abi)) {
+        const { decodedKey, decodedValue } = this.decodeAbi(key, value);
+        delete records[key];
+        records[decodedKey] = decodedValue;
```
sammyluo, line 176: So if the records of a domain for some reason is
```
{
  "abi.json": "random string"
}
```
it will become this?
```
{
  "abi.unknow": "random string"
}
```
Is that expected?

src/services/naming-services/EnsNamingService.test.ts
```diff
@@ -423,6 +423,83 @@ describe('EnsNamingService', () => {
...
+          },
+        },
+        records: {
+          contenthash:
+            '/ipfs/bafybeia4dgvuuh2k5kzu5a6neef7y2v6jtojz3ievehjp352zvcw4knorm',
+          'abi.json': [
+            'function store(string calldata mystr) public',
+            'function retrieve() public view returns (string memory)',
+          ],
+          'addr.0': 'bc1q4waj685749c7wueu8myavf2faz9gpl9802qg5t',
```
sammyluo, line 499: Can we also add some records that don't need to be decoded? It's to ensure that your algorithm doesn't delete or alter other values accidentally.

sammyluo (changes requested):
Good start!

src/models/Model.ts
```diff
@@ -27,7 +27,7 @@ import { Serialization, Serialized, Serializer } from '../services/Serializer';
 import { Attribute, Attributes, KeysOfType } from '../types/common';
 
 type ModelConstructor = {
-  [P in keyof typeof Model]: typeof Model[P];
+  [P in keyof typeof Model]: (typeof Model)[P];
```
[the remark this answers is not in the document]
tunguyennnnn: this is from prettier fix command

src/utils/qrImage.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
-import { createCanvas } from 'canvas';
+import { CanvasRenderingContext2D, createCanvas } from 'canvas';
```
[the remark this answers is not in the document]
tunguyennnnn: there's a global class https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D
but we should use the imported one

src/services/naming-services/EnsNamingService.ts
tunguyennnnn: it is handled, in the example https://www.npmjs.com/package/@ensdomains/content-hash it's only suggests to handle ipfs as special case

src/services/naming-services/EnsNamingService.ts
tunguyennnnn: it's already checks in `decodeRecords` with `else if (key.startsWith(EnsRecordType.Addr)) {`
this is to check if it's in the format of example `addr.0` or not

src/services/naming-services/EnsNamingService.ts
tunguyennnnn: yea, maybe we should just keep the original record here

tunguyennnnn: updated to return the original value

src/services/naming-services/EnsNamingService.test.ts
tunguyennnnn: updated, added another test for decode fail as well

src/services/naming-services/EnsNamingService.ts
sammyluo: I still think it's good to have a guard here in the function to make sure it won't break the logic when we reuse this it elsewhere.

tunguyennnnn: updated

sammyluo (approved):
LGTM!

